Issue #10, opened 2026-08-27
check_correlations() emits spurious 'NaNs produced' warnings from nnet


beckyfisher:
## Summary

`check_correlations()` emits "NaNs produced" warnings from inside `nnet` whenever
two factor predictors are perfectly or near-perfectly associated. The correlation
values it returns are correct; the warnings come from standard errors that
`check_correlations()` computes and then discards.

## Detail

`fill_factor_factor_correlations()` (`R/check-correlations.R`) takes the
deviance from a `summary()` of the fitted model:

```r
fit <- try(summary(nnet::multinom(dat.r[,var.1] ~ dat.r[,var.2], trace = FALSE))$deviance,
           silent = TRUE)
```

`summary.multinom()` computes the full variance-covariance matrix in order to
report standard errors. On a perfectly separated fit that matrix has negative
diagonal entries, and `sqrt(diag(vc))` warns. Only `$deviance` is used, so none
of that work is needed.

Perfect association is not an edge case here: it is guaranteed by
`factor.factor.interactions = TRUE`, which constructs pasted interaction columns
that are by definition perfectly predicted by their own components.

## Reproduction

```r
library(mgcv); library(FSSgam)

d <- case_study1
d$site  <- factor(d$site)
d$ZONE2 <- factor(ifelse(d$SCORE2 > median(d$SCORE2), "high", "low"))
d$ZONE3 <- factor(ifelse(d$complexity > median(d$complexity), "hi", "lo"))
tf <- gam(log.Herbivore.biomass ~ s(depth, k = 3, bs = "cr") + s(site, bs = "re"),
          data = d)

generate_model_set(use.dat = d, test.fit = tf, pred.vars.cont = "depth",
                   pred.vars.fact = c("ZONE", "ZONE2", "ZONE3"),
                   factor.factor.interactions = TRUE, max.predictors = 3,
                   null.terms = "s(site,bs='re')", k = 3)
#> Warning messages:
#> 1: In sqrt(diag(vc)) : NaNs produced
#> 2: In sqrt(diag(vc)) : NaNs produced
#> 3: In sqrt(diag(vc)) : NaNs produced
```

The returned correlation matrix is right: every interaction column correlates at
1.000 with its own components, and those models are excluded by `cov.cutoff`
exactly as intended. Only the warnings are spurious.

## Suggested resolution

Take the deviance from the fitted object rather than from its summary:

```r
[...]

Issue #9, opened 2026-08-27
fit_model_set() progress bar cannot be suppressed


beckyfisher:
## Summary

`fit_model_set()` always writes a `utils::txtProgressBar` to stdout. There is no
argument to turn it off, so every call -- including calls made inside a loop, a
Quarto chunk, or a test -- emits progress output that the caller cannot suppress
except by wrapping the call in `capture.output()`.

## Detail

Both fitting paths in `R/fit-model-set.R` open a progress bar unconditionally:

```r
pb <- utils::txtProgressBar(max = length(mod.formula), style = 3)
```

`check_correlations()` and `generate_model_set()` are silent, so the behaviour
is also inconsistent within the package.

Consequences:

- The published case studies loop over response variables, so the rendered
  vignettes carry one progress bar per taxon.
- The `@examples` blocks in `?fit_model_set` and `?full_subsets_gam` write
  progress output during `R CMD check`.
- The test suite has to route every call through a `capture.output()` wrapper
  (`fit_quietly()` in `tests/testthat/helper-fixtures.R`) to keep the reporter's
  own output legible.

## Suggested resolution

Add a `progress = interactive()` (or `quiet = FALSE`) argument to
`fit_model_set()`, forwarded from `full_subsets_gam()`, and guard the
`txtProgressBar`/`setTxtProgressBar`/`close` calls and the doSNOW
`.options.snow` progress callback with it. Defaulting to `interactive()` keeps
the bar for interactive use and removes it from scripts, vignettes and checks.

## Context

Found while building the test suite for #5. Not a correctness problem, so it was
left out of that work.

beckyfisher:
Fixed in 1.1.0, PR #17.

`fit_model_set()` and `full_subsets_gam()` gain `progress`, defaulting to
`interactive()`, so the bar appears at the console but not in scripts, reports
or `R CMD check`. With `progress = FALSE` no `txtProgressBar` is created and no
`.options.snow` callback is registered, so nothing reaches stdout at all.

Verified on the merged code, counting stdout lines from
`capture.output(res <- fit_model_set(...))`:

| `save.model.fits` | `progress = FALSE` | `progress = TRUE` |
|---|---|---|
| `TRUE` | 0 lines | 1 |
| `FALSE` | 0 lines | 1 |

The same commit removed a double `close(pb)` in the parallel branch of both
fitting helpers, which closed the bar inside the `if` and again after it.

[...]

Issue #8, opened 2026-08-27
Candidate model names depend on the session locale's collation


beckyfisher:
## Summary

Candidate model names are built with `sort()`, so they depend on the session's
`LC_COLLATE` setting. The same model set analysed on two machines can produce
different `modname` values and a different row order in `mod.data.out`.

## Detail

`build_model_formulas()` (`R/generate-model-set.R`) ends with

```r
names(mod.formula) = c("null", lapply(use.mods, FUN = paste, collapse = "+"))
```

and `use.mods` was produced by `enumerate_candidate_models()` as

```r
use.mods = unique(lapply(use.mods, FUN = sort))
```

`sort()` on a character vector uses the collation of the current locale. Under a
typical `en_US.UTF-8` locale, collation is case-insensitive; under `C` (which is
what `R CMD check` and testthat both set for reproducibility) it is by byte, so
uppercase sorts before lowercase.

```r
# en_US.UTF-8
"complexity+ZONE"
# C
"ZONE+complexity"
```

Both names describe the same model and the same formula, and the fitted results
are identical. What differs is:

- the `modname` column of `mod.data.out`, and hence anything a downstream script
  matches on by name;
- the order of `mod.formula`, and so the row order of `mod.data.out`;
- the labels in any figure built from `modname`.

The published case studies select the top model set by `delta.AICc` rather than
by name, so this does not affect them, but it does mean a saved analysis is not
byte-reproducible across machines whose locales differ.

## Suggested resolution

Make the ordering explicit rather than locale-dependent. `sort(method = "radix")`
sorts in the C locale regardless of `LC_COLLATE` and is the smallest change:

```r
use.mods = unique(lapply(use.mods, FUN = sort, method = "radix"))
```

Note this is a change in output for anyone currently running in a non-C locale:
existing saved model tables would no longer match on `modname`. It is worth
deciding whether that is acceptable now, before CRAN release, rather than after.

## Context

Found while building the test suite for #5 -- the new tests assert on `modname`
values, and had to be written against C collation because testthat forces it.
That is noted in a comment in `tests/testthat/test-generate_model_set.R`.

beckyfisher:
Fixed in 1.1.0, PR #17.

Candidate names are now built with `sort(method = "radix")`, which sorts in
byte order regardless of `LC_COLLATE`. Names are the C-locale form everywhere,
so a saved analysis is reproducible across machines.

Verified by generating the same model set under `LC_COLLATE=C` and
[...]

Issue #7, opened 2026-08-27
full_subsets_gam() does not forward VI.mods to fit_model_set()


beckyfisher:
## Summary

`full_subsets_gam()` accepts and forwards every other `fit_model_set()`
argument, but not `VI.mods`. The `VI.mods = 'all'` variable importance
calculation is therefore unreachable through the wrapper.

## Detail

`R/full-subsets-gam.R` forwards `max.models`, `save.model.fits`, `parallel`,
`n.cores`, `r2.type` and `report.unique.r2` to `fit_model_set()`. `VI.mods` is
absent from both the `full_subsets_gam()` formals and the forwarded call, so
`fit_model_set()`'s default `'min.n'` is always used.

```r
"VI.mods" %in% names(formals(fit_model_set))
#> [1] TRUE
"VI.mods" %in% names(formals(full_subsets_gam))
#> [1] FALSE
```

A user who wants importance scores summed over all models rather than over the
best n per predictor has to abandon `full_subsets_gam()` and call
`generate_model_set()` + `fit_model_set()` separately.

## Suggested resolution

Add `VI.mods = 'min.n'` to `full_subsets_gam()`'s formals and forward it, with a
matching `@param` entry copied from `fit_model_set()`. This is additive and does
not change any existing call's behaviour.

The same question applies to `full.subsets.gam()`, the deprecated alias: it
forwards `...` so it would pick the new argument up automatically.

## Context

Found while building the test suite for #5. The current behaviour is recorded by
a test in `tests/testthat/test-full_subsets_gam.R` ("full_subsets_gam has no
VI.mods argument and always uses the min.n default"), which will need updating as
part of the fix.

beckyfisher:
Fixed in 1.1.0, PR #17.

`full_subsets_gam()` gains `VI.mods`, defaulting to `'min.n'`, and forwards it
to `fit_model_set()`. An unrecognised value is rejected by `match.arg()` before
the candidate set is built rather than failing later with
`object 'aic.var.weights' not found`.

Verified on the merged code: `"VI.mods" %in% names(formals(full_subsets_gam))`
is `TRUE`.

One note for anyone writing a similar test. The obvious check — run the wrapper
with each setting and compare against `fit_model_set()` called directly —
passes even if nothing is forwarded, because on the standard fixture the two
settings give identical variable importance. Every model beyond the
`min.mods`-th carries a `wi.AICc` that rounds to 0.000 at the three decimal
[...]

PR #11, opened 2026-08-27
Build a comprehensive test suite before the refactor (#5)

### LINKED ISSUE #5: Build a comprehensive test suite before the refactor and CRAN submission
## Purpose

The package is about to be refactored and prepared for CRAN submission. The
existing `testthat` suite is too thin to serve as a regression harness for that
work: it confirms that the main entry points run and return objects of the
expected shape, but it does not pin down numerical results, and it leaves most
argument combinations, both non-default fitting paths, and the non-`gam` model
classes entirely unexercised. A refactor carried out against the current suite
could change fitted results, weights, or variable importance without any test
failing.

This issue covers building a comprehensive suite **before** the refactor
begins, so that the refactor can be verified rather than merely completed.

## Measured baseline

Run on 2026-08-27, branch `fix_issues`, R 4.6.1, via `covr::package_coverage()`:

```
FSSgam Coverage: 71.33%
R/fit-model-set.R:                 52.85%
R/generate-model-set.R:            66.37%
R/functions_supporting.R:          74.24%
R/check-correlations.R:            79.73%
R/full-subsets-gam.R:              91.67%
R/check-non-linear-correlations.R: 94.00%
R/utils.R:                         94.59%
R/deprecated.R:                    100.00%
```

The two core user-facing functions are the two least covered files. Re-measure
before starting, since this figure will move as work lands.

Line coverage overstates the position. The current suite is 780 lines across
eight files, against roughly 1,900 lines of R code, and most assertions are
structural (`expect_named()`, `expect_s3_class()`, dimensions, "no failed
models") rather than numerical. Three genuine behavioural regression tests
exist and should be preserved as they are: the two issue #10/#12 family
resolution tests, and the `full_subsets_gam()` `size=`/`used.data` tests from
the Phase 7 bug fixes.

## What is not tested at all

Confirmed by grep over `tests/` and by the zero-count lines reported by
`covr`. None of the following appears anywhere in the committed suite.

**`fit_model_set()` arguments and paths** (`R/fit-model-set.R`)

- `save.model.fits = FALSE`. The whole of `fit_and_summarise_unsaved_models()`
  (lines 196-245) is uncovered. This is a separate implementation, deliberately
  kept separate from the saved-models path so that fitted objects are never all
  held in memory at once (CLAUDE.md Section 6, Phase 6). It must be tested
  directly, and its `mod.data.out` must be shown to agree with the
  `save.model.fits = TRUE` path on the same model set.
- The `max.models` override (lines 91-96): exceeding `max.models` should emit a
  warning and silently force `save.model.fits = FALSE`. Neither the warning nor
  the downgrade is tested.
- `parallel = TRUE` for both fitting paths (lines 149-160 and 210-233). See
  "Constraints" below for how to handle this.
- `r2.type` — only `"r2"` and `"dev"` are touched, and only through a direct
  `extract_mod_dat()` call, never through `fit_model_set()`. The package default
  `"r2.lm.est"` is never asserted against a known value.
- `report.unique.r2 = TRUE` (lines 257-259): the `r2.vals.unique` column is
  never produced or checked.
- `VI.mods = 'all'` (lines 291-296). Only the `'min.n'` default is exercised,
  and even then `variable.importance` is only checked for presence in the
  returned list, never for value.

**`generate_model_set()` arguments** (`R/generate-model-set.R`)

- `cyclic.vars`: the whole `bs='cc'` substitution block (lines 578-588),
  including the `te()`-specific `bs=c(...)` construction for a cyclic variable
  inside a smooth-smooth interaction. `case_study3` has `month` and
  `lunar.date` and is the natural fixture.
- `linear.vars`: linear main effects and `.t.` linear interaction terms
  (lines 336-371 in `resolve_factor_interactions()`, line 583 in
  `build_model_formulas()`).
- `bs.arg` other than the `"'cr'"` default.
- A user-supplied `cor.matrix`, including the "supplied cor.matrix is missing
  required predictors" error path (lines 470-475).
- `factor.smooth.interactions` supplied as a list (per-factor control) rather
  than a character vector. The `NA` case is reached only incidentally, through
  the legacy `smooth.interactions` forwarding test in
  `test-full_subsets_gam.R`, and deserves a direct test against
  `generate_model_set()`.
- `smooth.smooth.interactions` combined with `non.linear.correlations = TRUE`
  (lines 424-451).
- The "max.predictors is greater than the number of predictors" error (line 492).
- `build_null_model()`'s failure path (lines 214-220), which raises the
  "Null model not successfully fitted" error, and its `null.terms = ""` branch.

**Model classes other than `mgcv::gam`**

`extract_mod_dat()`'s `gamm4`, `gamm` and `uGamm` branches (lines 70-104 of
`R/functions_supporting.R`), and `fit_mod_l()`'s `dsm` branch (lines 174-175),
are wholly uncovered, as are the `class(test.fit) == "gamm4"` branches in
`build_model_formulas()` that emit `t2()` in place of `te()`. `gamm4` is a
declared `Imports` dependency and `MuMIn::uGamm` is the documented way to fit
model sets with random effects. At minimum the `gamm4`/`uGamm` path needs
end-to-end coverage; the `dsm` path cannot be tested without adding a
dependency, so decide explicitly whether to leave it uncovered and record that
decision in a comment.

**`check_correlations()`**

The factor-factor correlation block and its `parallel = TRUE` branch
(lines 111-124, 135) are uncovered. The current test supplies a single factor,
so no factor-factor pair is ever formed.

**Datasets**

Only `case_study1` is used. `case_study2`, `case_study3` and `coral_data` are
exported, documented, and never loaded by any test. Each needs at least a
structural test (dimensions and column classes matching the `@format` entry in
`R/data.R`), and they should be used as fixtures for the cases they suit:
`case_study2` for factor-heavy and random-effect model sets, `case_study3` for
cyclic variables, `coral_data` for a wide predictor set.

## Numerical regression coverage

This is the part that matters most for the refactor, and is currently absent.
Structural assertions will not catch a refactor that quietly changes a fitted
value.

For a small number of representative, fast-fitting scenarios, pin the actual
numbers:

- the full sorted `mod.data.out` AICc, BIC, delta and weight columns;
- `variable.importance` for both `VI.mods` settings;
- the complete set of candidate model formulas produced by
  `generate_model_set()`, as a sorted character vector of `names(mod.formula)`
  plus the deparsed formulas.

Use `testthat::expect_snapshot_value()` (or `expect_snapshot()` on a printed
summary) so that the expected values live in `tests/testthat/_snaps/` and any
change is visible as a reviewable diff rather than an edited literal. Round
numeric values to a tolerance that is stable across platforms — mgcv's
smoothing-parameter optimisation is not bit-identical across BLAS
implementations, so a tolerance of roughly 1e-4 on AICc, and rounding weights
to three decimal places (which the package already does internally), is a
sensible starting point. Verify the snapshots are stable by running the suite
twice before committing.

Do not snapshot anything that embeds an absolute path, a timestamp, or an
environment address.

## Test infrastructure

The current files repeat the same eight-line `use.dat`/`test.fit`/`model.set`
setup roughly fifteen times. Before adding more tests, factor this out:

- Add `tests/testthat/helper-fixtures.R` with constructors returning the common
  fixtures, for example `fixture_cs1_gaussian()`, `fixture_cs1_tweedie()`,
  `fixture_cs2_gamm4()`, `fixture_cs3_cyclic()`. `helper-*.R` files are sourced
  automatically by testthat before the test files.
- Keep the fixtures deliberately small (`k = 3`, two or three predictors) so the
  whole suite stays fast. Time the suite before and after; report both figures
  in the pull request.
- Where a fixture must be visible to a `doSNOW` worker or must survive mgcv's
  normalisation of formula environments, follow the existing pattern in
  `test-fit_model_set.R` (assignment to the global environment with
  `on.exit()` cleanup) and keep the comment explaining why.

Suggested file layout, one file per source file, matching the existing
convention:

```
tests/testthat/helper-fixtures.R
tests/testthat/test-generate_model_set.R              (extend)
tests/testthat/test-generate_model_set_interactions.R (extend)
tests/testthat/test-generate_model_set_formulas.R     (new: cyclic.vars,
                                                       linear.vars, bs.arg,
                                                       gamm4 t2() branch)
tests/testthat/test-fit_model_set.R                   (extend)
tests/testthat/test-fit_model_set_options.R           (new: save.model.fits,
                                                       max.models, r2.type,
                                                       report.unique.r2,
                                                       VI.mods)
tests/testthat/test-fit_model_set_parallel.R          (new, skipped on CRAN)
tests/testthat/test-full_subsets_gam.R                (extend)
tests/testthat/test-check_correlations.R              (extend)
tests/testthat/test-check_non_linear_correlations.R   (extend)
tests/testthat/test-functions_supporting.R            (extend)
tests/testthat/test-utils.R                           (new)
tests/testthat/test-data.R                            (new)
tests/testthat/test-deprecated.R                      (keep as is)
```

## Constraints

These are established decisions in this repository. Read `CLAUDE.md` in full
before starting; the following are the ones most likely to be tripped over.

- **`s()`, `te()` and `t2()` must stay unqualified inside any `gam()`/`gamm()`
  formula.** `mgcv::s(...)` written inside a formula breaks `model.frame()`
  construction silently and surfaces as an unexplained `NA`. Namespace the
  outer `mgcv::gam()` call, never the smooth constructor inside the formula.
- **The three deprecated aliases (`generate.model.set()`, `fit.model.set()`,
  `full.subsets.gam()`) are permanent.** Their existing deprecation tests stay.
  Do not add aliases for the five names renamed without one (see CLAUDE.md
  Section 5).
- **Do not simplify `fit_mod_l()`'s family resolution.** It reconciles two
  directly conflicting issues (#10 and #12). Its two regression tests must
  continue to pass unchanged.
- **`parallel = TRUE` in committed tests.** The suite has deliberately avoided
  spinning up a real `doSNOW` cluster, because of CRAN check time and
  flakiness. The preferred approach here is a single, small
  `test-fit_model_set_parallel.R` guarded by `testthat::skip_on_cran()` and
  `testthat::skip_on_ci()` if it proves unstable, with `n.cores = 2`, so the
  path is covered locally without risking a CRAN check failure. Note that a
  `doSNOW` worker loads `FSSgam` from the installed library, never from
  `devtools::load_all()`, so `R CMD INSTALL .` must be current before the
  parallel test means anything — see CLAUDE.md Section 6, Phase 12.
- **Do not create a `vignettes/` folder.** Worked examples live in the
  companion publication repository, which is read-only for this work.
- **CRAN constraints apply to tests as well as to `R/`.** Keep the total suite
  runtime modest, use `skip_on_cran()` for anything slow, write no files outside
  `tempdir()`, and leave no objects behind in the global environment (the two
  existing `<<-` tests already clean up with `on.exit()`).
- The companion FAQ at
  https://github.com/beckyfisher/FSSgam/blob/master/vignettes/faq.Rmd
  and the case study vignettes in the same folder are the best source of
  realistic call patterns and documented edge-case behaviour. Read them before
  deciding what a given argument is supposed to do.

## Expected outcome

- Line coverage of at least 90% overall, with no file below 85%, or an explicit
  written justification in this issue for each exception (the `dsm` branch is
  the likely candidate).
- Every argument of `generate_model_set()`, `fit_model_set()` and
  `full_subsets_gam()` exercised at least once with a non-default value.
- Every documented error and warning message covered by a test asserting on the
  message text.
- Numerical snapshot tests in place for at least three end-to-end scenarios
  (Gaussian `gam`, an extended family such as `tw()` or `nb()`, and a
  `gamm4`/`uGamm` fit with a random effect).
- `devtools::check()` clean: 0 errors, 0 warnings, 0 notes.
- Suite runtime reported in the pull request, before and after.

Any bug found while writing these tests should be fixed in its own commit with
its own regression test, not folded into the test-suite work — this is the
precedent set in Phase 7, which is how the `size=` and `used.data` bugs in
`full_subsets_gam()` were handled.

## How to run

```r
devtools::load_all()
devtools::test()
devtools::check()
covr::package_coverage()          # per-file summary
covr::report(covr::package_coverage())   # line-level HTML, for finding gaps
```

`doSNOW` is a declared dependency but is not always installed in fresh
environments; it is required for `devtools::check()` and `load_all()` to
succeed.

### LINKED ISSUE #6: extract_mod_dat() returns NA r2 for gamm fits under the default r2.type
## Summary

`extract_mod_dat()` has no branch for a fitted object of class `gamm` (the class
returned by `MuMIn::uGamm(..., lme4 = FALSE)`, which wraps `mgcv::gamm`) under
either the package default `r2.type = "r2.lm.est"` or `r2.type = "dev"`. Both
silently return `NA` in `r2.vals`, for every model in the set.

## Detail

`R/functions_supporting.R` dispatches on `class(mod.fit)[[1]]`. The branches
present are:

| class | `r2` | `dev` | `r2.lm.est` |
|---|---|---|---|
| `gam` | yes | yes | yes |
| `gamm4` | yes | yes | yes |
| `gamm` | yes | **missing** | **missing** |

`tempOut` therefore stays `NA` and `mod.dat$r2.vals` is `round(NA, 5)`.

This matters because `r2.lm.est` is the package default, so the common case --
a user supplying a `uGamm(lme4 = FALSE)` test.fit and not setting `r2.type` --
produces a model table with an entirely `NA` `r2.vals` column, with no warning.
Setting `report.unique.r2 = TRUE` then propagates the `NA` to
`r2.vals.unique` as well.

`?fit_model_set` also documents `dev` as available for `gamm`:

> `dev` will return the deviance explained as reported by `gam` or `gamm`.

so the `dev` gap is a direct mismatch between documented and actual behaviour.

## Reproduction

```r
library(mgcv); library(MuMIn); library(FSSgam)

use.dat <- na.omit(coral_data[, c("Site", "av.wave", "Depth", "allcoral", "totalpoints")])
use.dat$Site <- factor(use.dat$Site)
use.dat$prop <- use.dat$allcoral / use.dat$totalpoints

test.fit <- uGamm(prop ~ s(Depth, k = 4, bs = "cr"),
                  random = list(Site = ~1), data = use.dat, lme4 = FALSE)
class(test.fit)
#> [1] "gamm"

extract_mod_dat(test.fit, r2.type. = "r2")$r2.vals
#> [1] 0.07883
extract_mod_dat(test.fit, r2.type. = "r2.lm.est")$r2.vals
#> [1] NA
extract_mod_dat(test.fit, r2.type. = "dev")$r2.vals
#> [1] NA
```

## Suggested resolution

Decide, and then make code and documentation agree:

- `dev` for `gamm`: `summary(mod.fit$gam)$dev.expl`, mirroring the `gamm4`
  branch (which already guards for a zero-length result).
- `r2.lm.est` for `gamm`: an observed-against-fitted regression is well defined
  here (`mod.fit$gam$y` against `predict(mod.fit$gam)`), but it needs a decision
  on whether the random effects are included in the prediction, which is the
  same decision the `gamm4` branch resolves with `re.form = NA`.
- If either is deliberately not supported, `extract_mod_dat()` should say so
  rather than returning `NA`, and the `?fit_model_set` `r2.type` documentation
  should be corrected.

## Context

Found while building the test suite for #5. The current behaviour is pinned by a
test in `tests/testthat/test-functions_supporting.R` ("extract_mod_dat handles a
gamm (nlme) fit") so that changing it is a visible diff rather than a silent
one; that test will need updating as part of the fix.

beckyfisher:
Closes #5.

## What this does

Builds the regression harness the refactor needs, before the refactor starts.

| | before | after |
|---|---|---|
| passing expectations | 105 | 447 |
| test files | 8 | 14 (+1 helper) |
| line coverage | 71.33% | 94.27% |
| suite runtime | 4.83 s | 15.5 s |

All measured on the same machine (WSL2, R 4.6.1) before and after: coverage with
`covr::package_coverage()`, expectations and runtime with
`testthat::test_dir()` in a warm session. Counts are testthat's own `PASS`
figure, which excludes skipped expectations. Runtime is machine-dependent;
re-measure rather than comparing against these figures. Note the baseline is 105
expectations across 36 `test_that()` blocks, measured on `master` — the "93
tests" figure in `CLAUDE.md`'s Phase 12 entry does not reproduce.

```
                             before    after
R/deprecated.R               100.00   100.00
R/utils.R                     94.59   100.00
R/full-subsets-gam.R          91.67   100.00
R/check-non-linear-corr...    94.00   100.00
R/generate-model-set.R        66.37    99.42
R/functions_supporting.R      74.24    95.45
R/check-correlations.R        79.73    81.08
R/fit-model-set.R             52.85    80.49
```

## Test infrastructure

`tests/testthat/helper-fixtures.R` replaces the eight-line
`use.dat`/`test.fit`/`model.set` preamble that was repeated in every file. It
provides data constructors for all four bundled datasets (prepared as the
companion repository's vignettes prepare them), `test.fit` constructors for
Gaussian/Tweedie/negative-binomial `gam`, a `case_study2` fit with nested
random effects, a cyclic Gamma fit, a `uGamm(lme4 = TRUE)` `gamm4` fit and a
`uGamm(lme4 = FALSE)` `gamm` fit, an overridable model-set constructor, and
`fit_quietly()`/`full_subsets_quietly()` wrappers that keep `fit_model_set()`'s
progress bar out of the reporter output.

Four groups of tests keep their original direct call form, because the issue
asks for them to be preserved as they stand: the two family resolution
regression tests, the two `full_subsets_gam()` Phase 7 regression tests, and
everything in `test-deprecated.R`.

## What is now covered

**`fit_model_set()`** — `save.model.fits = FALSE` (the whole separate
`fit_and_summarise_unsaved_models()` implementation, shown to produce a
`mod.data.out` identical to the saved path's on the same model set, compared in
full rather than column by column); the `max.models` warning and
its silent downgrade; all three `r2.type` values, each asserted against a value
recomputed from the corresponding fitted object; `report.unique.r2`; both
`VI.mods` settings, asserted by value; the "None of your models fitted
successfully" error; and partial failure, injected with a
`break_one_candidate()` helper.

**`generate_model_set()`** — `cyclic.vars` (main effects, `by` terms, and both
the all-cyclic and mixed `te()` `bs = c(...)` constructions); `linear.vars`
(main effects, `.t.` product terms, a predictor named in both
`pred.vars.cont` and `linear.vars`, and a linear-only model set); non-default
`bs.arg`; a user-supplied `cor.matrix` and its missing-predictor error;
`cov.cutoff`, asserted by which candidates it admits rather than merely passed;
`factor.smooth.interactions` as `NA`, as a character vector and as a list;
`factor.factor.interactions` and `smooth.smooth.interactions` in both their
logical and character-vector forms, including each form's own
exclusion-by-`cov.cutoff` pass, its `max.predictors` clamp, and every error it
raises; `smooth.smooth.interactions` with `non.linear.correlations = TRUE`; the
`max.predictors` error; `build_null_model()`'s failure path; and
`null.terms = ""`.

**Model classes other than `mgcv::gam`** — end-to-end `gamm4`/`uGamm` model
sets, `extract_mod_dat()`'s `gamm4` binomial and non-binomial `r2.lm.est`
branches, its `gamm` branch, and `fit_mod_l()` on a `gamm4` test.fit. The
`dsm` branch is left uncovered deliberately, with a comment in
`test-functions_supporting.R` explaining why: it needs a fitted `dsm` object,
and adding `dsm` plus the Distance/mrds stack as a dependency to cover two
lines of a plain `update()` call is not justified.

**`check_correlations()`** — the factor-factor `multinom` block, asserted
against a hand-computed pseudo-R2; the equal-deviance case, using two exactly
balanced factors; the factor-only path; single-column inputs.

**Datasets** — `test-data.R` checks all four against the dimensions, column
names and column classes their `@format` entries record.

**Numerical regression** — `test-numerical-regression.R` snapshots the full
sorted model table, both variable importance vectors, and the complete set of
deparsed candidate formulas, for five end-to-end scenarios: Gaussian `gam`,
negative binomial, binomial `gamm4`/`uGamm`, a cyclic `case_study3` set, and
`VI.mods = 'all'`.

The comparison is split in two, because `expect_snapshot_value()` passes its
tolerance to `waldo::compare()`, which applies it element-wise and *relatively*,
so one tolerance cannot serve both kinds of column. `delta.*`, `wi.*` and the
importance scores are already rounded to three decimal places by
`compute_model_weights()` and are compared at testthat's default tolerance,
which on a 0.001 grid is exact — a relative tolerance would reject a weight
moving from 0.007 to 0.0075 while accepting a change of 0.5 in a weight of 0.9.
`AICc`, `BIC`, `r2.vals` and `edf` are compared at 1e-6 relative, which on an
AICc of about 200 is 2e-4 absolute: far tighter than anything that could
reorder two models, and far looser than the scale on which BLAS differences
appear.

One quantity is excluded rather than tolerated: the `edf` of the binomial
`gamm4`/`uGamm` scenario. Measured on this branch's CI run at `7ec2b61`, which
executed all five scenarios with these tolerances in place, four passed in full
and the fifth differed only there — 4.000/5.000/8.000 locally against
4.260/5.050/8.350 on `ubuntu-latest`. gamm4 reports a smooth's `edf` through
lme4's random-effect machinery, so that column tracks the lme4 version rather
than anything this package does; it is asserted structurally in that scenario
instead. Everything else, in all five scenarios, is compared wherever the suite
runs. Snapshots were confirmed identical across three consecutive local runs.

## Bugs found and fixed

Each in its own commit with its own regression test, per the Phase 7 precedent
the issue cites.

1. **Single-predictor model sets failed.** `use.dat[, all.predictors]` dropped
   to a bare vector before `check_correlations()` /
   `check_non_linear_correlations()`, which classify the columns of a
   data.frame, so any model set with exactly one predictor died with an
   unrelated `argument is of length zero` (or, with
   `non.linear.correlations = TRUE`, `invalid 'row.names' length`) instead of
   building a 1x1 correlation matrix. Fixed with `drop = FALSE`, plus a
   zero-pair case in `check_non_linear_correlations()`.
2. **Phantom `NA.by.<factor>` terms.** With factors supplied and
   `pred.vars.cont = NA` — the documented way to run without smooth predictors
   — `expand.grid()` paired the `NA` itself with each factor. At
   `max.predictors = 1` the term was discarded again before the model set was
   returned, and the only symptom was `enumerate_candidate_models()` taking
   `max()` of an empty correlation sub-matrix and warning twice. From
   `max.predictors = 2` it survived into the candidate set, as a model whose
   formula smoothed the literal `NA`.
3. **Interaction terms dropped for every `linear.vars` entry after the first.**
   The pattern used to find `.t.` terms was built as
   `paste(linear.vars, ".t.")`, a vector when more than one linear predictor
   was supplied; `grep()` used only its first element. Affected candidates kept
   their interaction name but were fitted as the corresponding factor
   main-effect model, so the table reported duplicate fits under distinct
   names. Now matched with `startsWith()` against each linear predictor's own
   prefix, since predictor names routinely contain regex metacharacters.
4. **`case_study1` documented as 28 variables**; it has 27.
5. **Two `@return` blocks did not match what the functions return.**
   `generate_model_set()` documented a `generated.models` element that has never
   existed and omitted `n.mods`, `mod.formula`, `test.fit` and `included.vars`,
   which it does return. `fit_model_set()` claimed a `used.data` it has never
   returned — `full_subsets_gam()` is what carries that. Both reconciled, with
   `expect_named()` regression tests.
6. **A second route to the dropped-interaction defect in 3.** The `.t.` terms
   are generated from `factor.smooth.interactions$linear.vars` in the list form
   of that argument, which is validated against all the predictors rather than
   against `linear.vars`. Re-deriving the match from `linear.vars` therefore
   reopened the same silent drop whenever the two disagreed.
   `build_model_formulas()` now matches against the term names
   `resolve_factor_interactions()` actually generated.

## Issues raised rather than fixed here

Seven findings are behaviour decisions rather than defects, so they are recorded
as issues instead of being resolved in a test-suite pull request.

Two are pinned by an expectation that must change when the issue is fixed: #6
and #7. #8 is a partial case — candidate names are asserted throughout under C
collation, so any change to how they are built shows up, but the natural fix
(sorting by byte order) leaves those names exactly as they are. #9, #10 and #12
are recorded in comments or in `CLAUDE.md`, and fixing any of them would change
no test.

That asymmetry is deliberate: an expectation that fails when a defect is
corrected is worse than a comment. It is stated here so the comments are not
read as coverage.

- #6 `extract_mod_dat()` returns `NA` r2 for `gamm` fits under the package
  default `r2.type` (and for `"dev"`, which `?fit_model_set` documents as
  supported for `gamm`).
- #7 `full_subsets_gam()` does not forward `VI.mods`.
- #8 Candidate model names depend on the session locale's collation.
- #9 `fit_model_set()`'s progress bar cannot be suppressed.
- #10 `check_correlations()` emits spurious `NaNs produced` warnings out of
  `nnet`.
- #12 The factor-factor correlation diagonal is not exactly 1.
- #13 A user-supplied `cor.matrix` does not govern `te()` term construction.

Note that the `#10` and `#12` referred to throughout `R/` and `CLAUDE.md` for
the family resolution work are issues in the **publication** repository,
`beckyfisher/FSSgam`, not in this one. References added by this branch are
qualified explicitly (`beckyfisher/FSSgam#10`, `FSSgam_package#6`), and
`CLAUDE.md` Section 6 now records the convention.

## Parallel tests

`test-fit_model_set_parallel.R` covers both fitting paths, the extended-family
case and the `family = family.opts[[2]]` case under `parallel = TRUE`, plus
`full_subsets_gam()`'s forwarding. Every test is guarded by
`skip_unless_parallel_opt_in()`, which requires `FSSGAM_TEST_PARALLEL=true` and
an installed (not pkgload) copy.

Two reasons for the opt-in. A doSNOW worker always loads the *installed*
package, so these mean nothing against a `devtools::load_all()` copy. And
doSNOW cluster startup stalls indefinitely on the WSL host used for this work,
including on an idle machine: `test-fit_model_set_parallel.R` completed cleanly
on two of four consecutive attempts and stalled past 180 s on the other two,
while the same code run from a plain script completed six times out of six. An
unattended stall would consume the entire runtime of a check job, and
`skip_on_cran()` alone would not prevent it, because `devtools::check()` sets
`NOT_CRAN=true`.

`.github/workflows/parallel-tests.yaml` is the one place they run
automatically. It installs the package, sets the variable, is scoped with
`filter =` to the two files carrying parallel tests, carries its own
`timeout-minutes`, and fails if any of those tests is skipped.

Measured on `ubuntu-latest`: **all 14 expectations pass, with no stall.** The
job itself was red on two earlier commits of this branch — once on a snapshot
comparison and once because its skip assertion counted the snapshot skips added
in the same commit — both of which are fixed here. The instability is specific
to the local WSL host, not to the package or to Linux.

Locally:

```
R CMD INSTALL .
FSSGAM_TEST_PARALLEL=true NOT_CRAN=true Rscript -e \
  'library(FSSgam); testthat::test_dir("tests/testthat", package = "FSSgam")'
```

## Coverage justification

The issue asks for 90% overall and 85% per file, or an explicit written
justification per exception. Overall is 94.27%. Two files fall below 85%, and
in both the entire shortfall is the `parallel == TRUE` branches, which `covr`
does not execute because of the opt-in above:

- `R/check-correlations.R` 81.08% — uncovered lines 111-124, the whole
  `parallel == TRUE` branch of `fill_factor_factor_correlations()`. With those
  executed the file reaches 100%.
- `R/fit-model-set.R` 80.49% — uncovered lines 149-160 and 213-226, the
  `parallel == TRUE` branches of the two fitting helpers. With those executed
  the file reaches 100%.

With the parallel branches counted, overall coverage is about 99.3%. Codecov
reports the same 94.27%.

Everything else uncovered is deliberate and enumerated:

- `R/generate-model-set.R:213` and `R/functions_supporting.R:174-175` — the
  `dsm` branches, which need a `dsm` fit and therefore the Distance/mrds
  dependency stack.
- `R/generate-model-set.R:257` — unreachable. It is a
  `length(pred.vars.fact) < 2` guard nested inside a
  `length(na.omit(pred.vars.fact)) > 1` branch.
- `R/functions_supporting.R:87` — `if (is.null(tempOut)) tempOut <- NA`, a
  defensive guard for a `summary()` field that no currently supported model
  class returns as `NULL`.

## Declared dependency floors

`DESCRIPTION` declares `R (>= 3.5)`, so the suite must run on that. The tests
use a `deparse_one()` helper rather than `deparse1()`, which arrived in R 4.0.0.
The R floor itself is deliberately unchanged: `deparse1()` was only ever a test
convenience, and raising the package's declared R dependency to buy one is not
justified.

The testthat floor is a different matter and **is** raised here, from
`>= 3.0.0` to `>= 3.1.5`: `expect_no_warning()` arrived in 3.1.5 and the suite
uses it in five places. testthat is a `Suggests` used only by the tests, so the
floor should follow what they need.

## Mutation testing

The branch was reviewed in nine rounds. Six of them (rounds 3 to 8) applied
single-line mutations to `R/` and re-ran the whole suite against each, at
9, 9, 32, 37, 37 and 27 mutations respectively; a further 8 were run to verify
individual fixes. That is 159 runs rather than 159 distinct mutations, since the
rounds re-tested each other's findings. Every mutation of
the fitting statistics, the weight and variable-importance arithmetic, the
formula construction, the correlation estimators, the family resolution and
`full_subsets_gam()`'s argument forwarding fails at least one expectation.

Eleven survivors were real gaps, each now closed by a test written for it:

- three `full_subsets_gam()` forwarding arguments — `bs.arg`, `cyclic.vars` and
  `factor.factor.interactions` — none of which leaves a trace in the candidate
  *names*, which is all the tests named after them had been comparing;
- the line putting generated `.I.` columns into `all.predictors` before the
  collinearity screen, without which a candidate could pair a pasted
  interaction with one of the factors it is built from;
- the lower-triangle collinearity test, in both places it appears. Factor-factor
  correlations are asymmetric, so which triangle carries an exceedance depends
  on the order the predictors were named;
- `max.predictors` counting variables rather than terms, so that a `te()` term
  consumes two of them;
- the pairwise-only restriction on `te()` terms;
- the `.I.` counterpart of that count, so a pasted factor-factor interaction
  column counts as its two component variables;
- `enumerate_candidate_models()`'s deduplication, and `extract_mod_dat()`'s
  `edf.less.1` threshold, both of which had assertions that did not in fact
  constrain them.

Seven survivors cannot be caught by any test and are recorded rather than
chased: `>` to `>=` on `cov.cutoff`; an `nchar(null.terms) == 0` guard
overwritten by the branch below it; `environment(formula(test.fit))` to
`parent.frame()`, which `CLAUDE.md` Phase 12 already records as unpinnable
because mgcv normalises the formula environment away; a
`cor.matrix[which(cor.matrix == "NaN")] <- 0` line redundant with the
`is.na()` assignment beneath it; two `max.predictors` clamps made dead by the
loop guards immediately below them; and the `cov.cutoff` screen inside
`resolve_smooth_smooth_interactions()`, which duplicates one
`enumerate_candidate_models()` applies afterwards. That last one is separable
in exactly one configuration — a supplied `cor.matrix`, which governs candidate
exclusion but not `te()` term construction — and that asymmetry is raised as
#13 rather than changed here.

## Checks

- `testthat::test_dir()`: `[ FAIL 0 | WARN 0 | SKIP 6 | PASS 447 ]` across 166
  `test_that()` blocks, 15.3 s. The 6 skips are the parallel tests.
- `devtools::check()`: **0 errors, 0 warnings, 0 notes** (46.1 s). An earlier
  `'::' or ':::' import not declared from: 'lme4'` warning was found and fixed
  during this work — the gamm4 test reads the `merMod`'s `flist` slot rather
  than calling `lme4::ranef()`, since lme4 reaches the tests only as a
  dependency of gamm4.


beckyfisher:
## Review summary

This branch was developed and reviewed over nine cycles. Each cycle consisted of
an independent session reviewing the branch with no prior context, followed by a
cycle addressing what it found. The last three found no blocking issues; the
ninth found nothing requiring change and recommended merging.

### Final state

| | before | after |
|---|---|---|
| passing expectations | 105 | 447 |
| `test_that()` blocks | 36 | 166 |
| line coverage | 71.33% | 94.27% |
| suite runtime | 3.6 s | 14.6 s |

`devtools::check()`: 0 errors, 0 warnings, 0 notes. All four workflows green on
ubuntu-release, ubuntu-devel, macOS and Windows, including the new
`parallel-tests` job, which reports `skipped: 0`.

### What the review rounds found

The reviews were more productive than the summary above suggests. They are
recorded here because the failures are the useful part.

**Seven source defects**, six of them found by review rather than by the tests
that were being written:

1. Single-predictor model sets failed with an unrelated error, because
   `use.dat[, all.predictors]` dropped to a vector.
2. A phantom `NA.by.<factor>` term was built when `pred.vars.cont = NA`. First
   recorded as harmless; it is not — from `max.predictors = 2` it survives into
   the candidate set as a model that smooths the literal `NA`.
3. Interaction terms were silently dropped for every `linear.vars` predictor
   after the first, so the model table reported duplicate fits under distinct
   names.
4. A second route to the same defect, through the list form of
   `factor.smooth.interactions`, which the first fix reopened.
5. `case_study1` was documented with 28 variables; it has 27.
6. `generate_model_set()`'s `@return` documented an element that has never
   existed and omitted three that do.
7. `fit_model_set()`'s `@return` claimed a `used.data` it has never returned.

**Eleven test gaps found by mutation testing**, each closed by a test written
for it. Several were assertions that looked like coverage and were not: a
deduplication check placed where deduplication never fires, an `edf` threshold
fixture that separated the wrong pair of values, and three
`full_subsets_gam()` forwarding tests that passed with the argument they were
named after hard-coded away.

**Two measurement errors in the records.** Every expectation count reported for
this work was six too high, because the figure came from a metric that counts
skipped expectations. And the count went out of step between `NEWS.md`,
`CLAUDE.md` and this description in five consecutive rounds. `CLAUDE.md` now
records the rule that follows rather than just the corrected number.

### Deliberate limits

- The `parallel = TRUE` tests are opt-in behind `FSSGAM_TEST_PARALLEL`. doSNOW
  cluster startup stalls indefinitely on the development machine used here —
  reproducibly, including for a trivial `foreach()` containing no package code —
  and an unattended stall consumes a whole check run. They run in their own CI
  job, which is green.
- Coverage is 94.27% rather than higher because `covr` does not execute those
  branches. Every uncovered line in `check-correlations.R` and
  `fit-model-set.R` is inside a `parallel == TRUE` branch; with them counted,
  overall coverage is about 99.3%.
- The numerical snapshots exclude one quantity, the `edf` of the `gamm4`
  scenario, which tracks the lme4 version rather than anything this package
  does.
- Seven mutation survivors cannot be caught by any test and are enumerated in
  the description rather than chased.

### Issues raised rather than fixed

Eight findings are behaviour decisions rather than defects, so they are recorded
as issues instead of being resolved in a test-suite pull request: #6, #7, #8,
#9, #10, #12 and #13. Two are pinned by an expectation that must change when the
issue is fixed; the rest are recorded in comments, which the description says
explicitly so they are not read as coverage.

### Recommendation

Ready to merge. Nothing outstanding affects package behaviour.

Issue #16, opened 2026-09-01
check_correlations compares deviances computed on different row sets when a factor has missing values


beckyfisher:
## What

In `check_correlations()`, the factor-factor correlation between two factors is
`sqrt(1 - fit/null)`, where the two deviances are computed on different sets of
rows when either factor contains missing values. The pair model is fitted on the
pair's complete cases; the null model is fitted on the whole column.

## Why it matters

The two deviances are not comparable when they come from different row sets, so
`r.est` for that pair is wrong by an amount that depends on how much of the
column is missing. There is no warning.

It is reachable only through a direct call to `check_correlations()`, because
`generate_model_set()` rejects predictors containing `NA` before it gets here.
`check_correlations()` is exported and documented as taking any data frame, so a
direct call is a supported use.

## Evidence

`R/check-correlations.R`, in `fill_factor_factor_correlations()`:

```r
dat.r <- stats::na.omit(dat[, c(var.1, var.2)])
fit   <- try(nnet::multinom(dat.r[, var.1] ~ dat.r[, var.2], trace = FALSE)$deviance, ...)
```

against the null, which is fitted once per factor on `dat`:

```r
null.deviances <- lapply(fact.vars, function(v) {
  try(nnet::multinom(dat[, v] ~ 1, trace = FALSE)$deviance, silent = TRUE)})
```

`dat.r` has had incomplete cases for that pair removed; `dat` has not. This is
long-standing, not introduced by the change that hoisted the null fit out of the
loop — before that hoist the null was already fitted on `dat` inside the loop.

<details>
<summary>Implementation detail</summary>

### The choice

Either deviance can be moved to match the other:

1. **Fit the null on the pair's complete cases.** Correct for the ratio, but
   returns the null fit to being per-pair rather than per-factor, undoing the
   reduction from n(n-1) fits to n. For n factors that is n(n-1) null fits
   again.
2. **Fit the pair on the whole column.** Not possible: `multinom()` needs
   complete cases for both variables.
3. **Drop incomplete cases once, across all the factor columns, before the
   loop.** One row set for everything, so every pair is comparable with every
   other as well. Changes the estimate for every pair when any factor has
   missing values, and discards more data than any single pair requires.

Option 3 is the most defensible statistically and the largest behaviour change.

### Impact

Any pair where either factor carries `NA`. Reported values change; whether they
move up or down depends on the data. Anything that stores a correlation matrix
from `check_correlations()` on data with missing factor values would need
[...]

Issue #15, opened 2026-09-01
A supplied cor.matrix must name factor interaction columns the user cannot know in advance


beckyfisher:
## What

With `factor.factor.interactions` set, a user-supplied `cor.matrix` must now
contain a row and a column for every hard coded interaction column the setting
causes to be created, or `generate_model_set()` stops. Supplying one for three
factors at `max.predictors = 3` means naming four extra columns
(`a.I.b`, `a.I.c`, `b.I.c`, `a.I.b.I.c`) that do not exist in `use.dat` and that
the user has to construct the names of by hand.

## Why it matters

This is the documented contract — `?generate_model_set` has always said a
supplied matrix "must include these hard coded interactions" — and it became
load-bearing rather than aspirational when FSSgam_package#13 made the supplied
matrix govern that stage. But it is onerous, and the failure mode is a stop
partway through model set construction rather than anything a user could
anticipate from their own data.

The names are deterministic (`paste(combn(factors, i), collapse = ".I.")`), so
this is a usability problem rather than an impossibility.

## Evidence

On this branch, with `ZONE` and `ZONE.copy` as factors and a supplied 3x3 matrix
covering `depth`, `ZONE` and `ZONE.copy`:

```
Error in build_predictor_correlation_matrix(...) :
  Supplied cor.matrix is missing required predictors: ZONE.I.ZONE.copy.
```

The matrix has to be built after working out which interactions will survive the
screen, which depends on the matrix.

<details>
<summary>Implementation detail</summary>

### Where

`resolve_factor_interactions()` in `R/generate-model-set.R` screens the
factor-factor combinations against the supplied matrix via its local
`factor_correlations()`, and `build_predictor_correlation_matrix()` then
validates the same matrix against `all.predictors`, which by that point includes
the interaction columns just created.

### Options, none implemented

1. **Fall back per predictor.** Use the supplied value where the matrix has one
   and compute the rest from `use.dat`. Keeps the user's intent for the
   predictors they named without requiring them to anticipate derived columns.
   Needs a decision about what "the rest" means when the supplied matrix and the
   computed one disagree about the same pair.
2. **Compute the interaction rows.** Accept a matrix covering only the named
   predictors, and extend it with computed rows and columns for any interaction
   column built. Narrower than option 1 and probably what a user expects.
3. **Document the construction.** Leave the behaviour and give a worked example
   in `?generate_model_set` showing how to build the required names. Cheapest,
   and does nothing about the error arriving late.

[...]

Issue #13, opened 2026-08-27
A user-supplied cor.matrix does not govern te() term construction


beckyfisher:
## Summary

A user-supplied `cor.matrix` governs which candidate models survive the
collinearity screen, but not which `te()` smooth-smooth interaction terms are
constructed in the first place. The construction stage recomputes correlations
from `use.dat` regardless of what was supplied.

## Detail

`generate_model_set()` screens for collinearity in two places.

`resolve_smooth_smooth_interactions()` (`R/generate-model-set.R`) decides which
pairs of continuous predictors become `te()` terms. It calls
`check_correlations()` or `check_non_linear_correlations()` on `use.dat`
directly:

```r
if(non.linear.correlations==TRUE){
 continuous.correlations=check_non_linear_correlations(use.dat[,pred.vars.cont])}else{
 continuous.correlations=check_correlations(use.dat[,pred.vars.cont])}
```

`build_predictor_correlation_matrix()`, which runs afterwards, is the function
that honours a supplied `cor.matrix`, and its result is what
`enumerate_candidate_models()` screens the assembled candidates against.

So a supplied `cor.matrix` cannot admit a `te()` term that the recomputed
correlations exclude, and cannot exclude one they admit — the term is built or
not before the supplied matrix is consulted.

`?generate_model_set` documents `cor.matrix` as governing exclusion generally:

> Note that any user constructed pairwise matrix can be passed to the function
> and used for pairwise exclusion of variables from individual models.

## Reproduction

```r
library(mgcv); library(FSSgam)

use.dat <- case_study1
use.dat$site <- as.factor(use.dat$site)
test.fit <- gam(log.Herbivore.biomass ~ s(depth, k = 3, bs = "cr") + s(site, bs = "re"),
                data = use.dat)

pred.vars <- c("depth", "SCORE2", "complexity")
zero.cor  <- matrix(0, 3, 3, dimnames = list(pred.vars, pred.vars))
diag(zero.cor) <- 1

model.set <- generate_model_set(
  use.dat = use.dat, test.fit = test.fit,
  pred.vars.cont = pred.vars, smooth.smooth.interactions = TRUE,
  cor.matrix = zero.cor, null.terms = "s(site,bs='re')",
  max.predictors = 2, k = 3
)
grep(".te.", names(model.set$mod.formula), fixed = TRUE, value = TRUE)
```

`depth.te.complexity` is absent, although the supplied matrix records those two
[...]

Issue #12, opened 2026-08-27
check_correlations() factor-factor diagonal is not exactly 1


beckyfisher:
## Summary

`check_correlations()` returns a factor-factor diagonal of approximately
0.999999 rather than exactly 1, because that value is a fitted
`nnet::multinom()` pseudo-R2 rather than a set constant. The
continuous-continuous diagonal, coming from `cor()`, is exactly 1, and
`check_non_linear_correlations()` sets its diagonal explicitly.

## Detail

`fill_factor_factor_correlations()` (`R/check-correlations.R`) fills every cell
of the factor-factor block from a multinomial fit, including the cells where
`var.1 == var.2`. The self-pair fit converges to a deviance near zero rather
than exactly zero, so `sqrt(1 - fit/null.fit)` lands just below 1.

```r
library(FSSgam)
d <- case_study1
d$ZONE <- factor(d$ZONE)

check_correlations(d[, c("depth", "ZONE")])
#>           depth      ZONE
#> depth 1.0000000 0.0823...
#> ZONE  0.0823... 0.9999994
```

Three consequences:

- The value depends on `nnet`'s optimiser tolerance, not on the data, so it is
  not reproducible in the way the rest of the matrix is.
- A user comparing the returned matrix against an expected one, or checking
  `all(diag(cm) == 1)`, gets an unexplained near-miss.
- `check_correlations()` and `check_non_linear_correlations()` disagree on the
  same input for no reason a caller can see.

It does not affect model selection: `enumerate_candidate_models()` reads only
the off-diagonal entries via `upper.tri()`/`lower.tri()`.

## Suggested resolution

Set the diagonal explicitly after the factor-factor block is filled, as
`assemble_non_linear_correlation_matrix()` already does:

```r
diag(out.cor.mat) <- 1
```

Skipping the self-pairs in the `lm.grid` loop would also avoid
`length(fact.vars)` redundant `multinom()` fits per call.

## Context

Found while building the test suite for #5. The tests in
`tests/testthat/test-check_correlations.R` currently assert
`all(diag(cm) > 0.99)` with a comment, rather than equality to 1, precisely
because equality would pin `nnet`'s optimiser rather than this package's
behaviour. They should become exact assertions when this is fixed.

beckyfisher:
Fixed in 1.1.0, PR #17.

`fill_factor_factor_correlations()` enumerated all ordered pairs of factors,
including each factor with itself. Regressing a factor on itself gives a
[...]

PR #17, opened 2026-09-01
Pre-CRAN refactor: fix known defects, restructure interaction resolution, bump to 1.1.0

beckyfisher:
## What

Fixes the defects and unvalidated arguments found in a full read of `R/`,
restructures the interaction-resolution code, and bumps the version to 1.1.0.
One commit per item, each behaviour fix carrying its regression test in the
same commit.

Seven phases, in the order set out in `planning/pre-cran-refactor-human.md`:
defects and validation, the correlation functions, the restructure (no
behaviour change), the behaviour changes, idioms, unexporting `fit_mod_l()`,
and documentation.

## Why it matters

`R CMD check` was already clean and stayed clean throughout, so none of this
was required to pass it. That is the argument for doing it now rather than
after release: the check does not call `generate_model_set()` with two
predictors whose names overlap, or with a single predictor, or with
`max.predictors = 1`, or with an invalid `r2.type`. Each of those was broken
and each passed the check.

Three of the changes alter output or the returned object for existing users,
which is why the version is 1.1.0 and why this lands before the first CRAN
submission rather than after: releasing first would make all three breaking
changes to a published package.

## What changes for a user

- **Candidate model names are sorted in byte order.** A model named
  `complexity+ZONE` in an `en_US.UTF-8` session was named `ZONE+complexity` in
  a C-locale one. It is now the C-locale form everywhere. Fitted results are
  identical; `modname` values and the row order of `mod.data.out` change for
  anyone in a non-C locale, and a saved model table matched on `modname` needs
  regenerating.
- **A supplied `cor.matrix` now governs every correlation screen**, not only
  which assembled candidates survive. It also decides which factor-factor
  interaction columns and which `te()` terms are built, and it now replaces the
  automatic estimate rather than overriding it, so `check_correlations()` is no
  longer called at all when a matrix is supplied. Users who do not supply one
  see no change.
- **`full_subsets_gam()`'s `max.models` default changes from 500 to 200**,
  matching `fit_model_set()`. The two disagreed and neither documented its
  default. A candidate set of 201 to 500 models now returns an empty
  `success.models` and a "model fits will not be saved" warning through the
  wrapper, as it already did through `generate_model_set()` plus
  `fit_model_set()`. Pass `max.models = 500` to keep the old behaviour.
- **Invalid argument values are errors.** `r2.type = "rsq"` returned a column
  of `NA` with no message; `VI.mods = "alll"` failed with
  `object 'aic.var.weights' not found`.
- **`fit_mod_l()` is no longer exported**, remaining available as
  `FSSgam:::fit_mod_l()`. `wi()`, `extract_mod_dat()` and
  `build_inclusion_mat()` stay exported.
- **New arguments:** `progress` on `fit_model_set()` and `full_subsets_gam()`,
  defaulting to `interactive()`; and `VI.mods` on `full_subsets_gam()`, which
  previously could not reach it.

`NEWS.md` describes each change and what to do about it.

## Evidence

| | `master` | this branch |
|---|---|---|
| `testthat` | 447 passing, 0 failing, 6 skipped | 568 passing, 0 failing, 7 skipped |
| line coverage (`covr`) | 94.27% | 94.39% |
| `devtools::check()` | 0 errors, 0 warnings, 0 notes | unchanged |
| `R CMD check --as-cran`, incoming checks on | Status OK | Status OK |

Measured on Debian (WSL2), R 4.6.1, testthat 3.3.2, on a clean checkout with
`tests/testthat/_snaps/` present, quoting `sum(as.data.frame(res)$passed)` and
`sum(...$skipped)` from `testthat::test_local()`. Skipped counts are
`test_that()` blocks: seven `parallel = TRUE` blocks that require
`FSSGAM_TEST_PARALLEL=true` and an installed package. Regenerating the snapshot
file skips five further blocks and the nine expectations inside them, which is
where an earlier, lower figure quoted on this pull request came from.

Every behaviour change was confirmed to fail before its fix, with one stated
exception: FSSgam_package#10, where the warning being removed is not
reproducible on nnet 7.3-20. That is written on the face of the test so it is
not mistaken for a regression test.

Phases 3 to 5 were checked against a golden master of 31 `generate_model_set()`
scenarios compared with `identical()`, capturing formulas, correlations,
column names, warnings and errors. Phase 2 used a separate 14-scenario master
over the two correlation functions. The scripts are in
`planning/golden-master/`. The review round below was checked the same way:
31 of 31 scenarios identical. Its six regression tests were each run against
the branch as it stood before the review, where all six fail.

<details>

### The four defects

- **A1** `cyclic.vars` was matched into assembled term strings with an
  unanchored `grep()`, so declaring `depth` cyclic also fitted `depthx` with
  `bs = "cc"`, and a name containing a full stop matched any character. The
  model was fitted with the wrong basis, silently, under a candidate name that
  looked correct. The basis is now chosen from the variable names as each term
  is built, which removed both post-hoc rewrite blocks.
- **A2** `fit_model_set()` failed on a single-predictor model set with
  `'x' must be an array of at least two dimensions`.
- **A3** `for(i in 2:n)` counts backwards at `n = 1`. On a two-factor set at
  `max.predictors = 1`, `used.data` had 35 columns with 34 duplicated names;
  it now has 32 with none.
- **A4** `.errorhandling='pass'` was commented out on the unsaved parallel
  path, so one unfittable candidate aborted the whole run.

### Defects found while doing the work, not in the original review

- The `te()` repair block took only the first two variables of a term, so a
  three-way `te()` was given two `bs` values. mgcv warns
  `bs wrong length and ignored` and substitutes its default, discarding both
  `bs.arg` and any `cyclic.vars` for that term.
- The character `factor.factor.interactions` branch reached
  `cbind(use.dat, <0-column data.frame>)` when the screen left nothing, failing
  with `arguments imply differing number of rows`, where the logical branch
  warned.
- An outer guard on the same branch counted cells of the correlation matrix
  rather than pairs, and skipped the whole block in silence.
- The 16 `@param` blocks duplicated between `generate_model_set()` and
  `full_subsets_gam()` drifted during this work, when `cor.matrix` was updated
  in one file only. They are now inherited.

### Defects found in review of this branch

- `build_predictor_correlation_matrix()` computed `check_correlations()` (or
  `check_non_linear_correlations()`) over every predictor and discarded the
  result whenever a `cor.matrix` was supplied, so the headline change above did
  not hold where it matters most: the fits still ran, and a predictor of a class
  those functions reject — a `Date` column, say — still aborted the call even
  though its correlations had been supplied.
- With `save.model.fits = FALSE`, a run in which no candidate fitted returned a
  table of `NA` with `delta.AICc` and `wi.AICc` as `NaN`, rather than the
  "None of your models fitted successfully" error the other path raises. Which
  of the two a user got depended only on a memory setting.
- `full_subsets_gam()` validated `r2.type` at entry but not `VI.mods`, and
  `progress` had no validation at all.
- `smooth.smooth.interactions` naming a column of `use.dat` that is not a
  predictor was reported as "not supplied in use.dat", which is not what the
  check tests.
- `man/FSSgam-package.Rd` had not been regenerated after `DESCRIPTION`'s
  Description field was rewritten.

### The restructure

Four copies of "enumerate every combination, drop the ones that are too
correlated" became `combine_uncorrelated()` and `combination_sizes()`. The two
`factor.smooth.interactions` forms are normalised once. Both smooth-smooth
copies built a data frame of pasted columns that was never read; it is gone.

Four divergences between the duplicated branches were held at their existing
behaviour through phase 3 behind explicit arguments, then resolved in phase 4,
so the restructure is verifiable as a pure no-op. Two are now fixed (both
triangles of the correlation matrix are screened everywhere; both branches warn
rather than one erroring), one is removed (the cell-counting guard), and one is
retained by decision and documented (`smooth.smooth.interactions = TRUE` builds
bivariate `te()` terms while a character vector builds terms up to
`max.predictors`).

A3 and the `cbind` divergence are committed together, because fixing A3 made
the erroring path reachable; committing either alone would have left the tree
broken.

### Issues

Fixed here: FSSgam_package#7, #8, #9, #10, #12, #13.

Raised and not fixed here: FSSgam_package#14 (`parallel = TRUE` stalls when
`gamm4` is loaded on the doSNOW workers — not a defect in this package, a
`foreach()` with a body of `i * 2` reproduces it, and `master` stalls at the
same rate), #15 (a supplied `cor.matrix` must name factor interaction columns
the user cannot know in advance), #16 (`check_correlations()` compares
deviances computed on different row sets when a factor has missing values).

Unchanged and still open: FSSgam_package#6.

### The parallel tests

Both fitting paths were verified against the final code, one cluster per
process: `AICc` matches the sequential result on both, with the injected
failing candidate recorded rather than aborting the run.

`test-fit_model_set_parallel.R` itself could not be run to completion on this
host. It creates about seven clusters, and at the per-cluster stall rate of
FSSgam_package#14 the chance of all seven completing is about one per cent;
measured 10 stalls in 10 attempts. Do not read a whole-file stall as a failure
of the code under test, and expect
`.github/workflows/parallel-tests.yaml` to need retry logic on any runner that
reproduces it.

### Review notes

`planning/` holds the two planning documents and the golden-master scripts, and
is in `.Rbuildignore`, so none of it reaches the built package. The
`planning/golden-master/README.md` explains why four of the 31 scenarios exist:
the obvious scenarios do not discriminate between the duplicated branches, and
each of the four was constructed against a measured correlation value.

Four files in `R/` use CRLF line endings — `data.R`, `FSSgam.R`,
`full-subsets-gam.R` and `functions_supporting.R`. An editor that rewrites them
as LF turns a three-line change into a whole-file diff; the diffs here are
clean.

`CLAUDE.md` gains a Phase 14 entry recording what a later session needs from
this work.

</details>

https://claude.ai/code/session_014UzX8JZfS6YUUrsLNe6XA4


beckyfisher:
## Review round: what changed on the branch

Five review cycles over the branch. Five commits, `276c2f5`..`f9b1921`. Each of
the five changes to the package carries a regression test, and all six of those
tests were run against the branch as it stood before the review (`0666ca6`),
where all six fail. The 31-scenario golden master is identical across the round,
so nothing moved that was not the point of a change. `R CMD check` stays at 0
errors, 0 warnings, 0 notes.

**A supplied `cor.matrix` did not replace the automatic estimate, only override
it.** `build_predictor_correlation_matrix()` computed `check_correlations()` (or
`check_non_linear_correlations()`) over every predictor and then discarded the
result whenever a matrix had been supplied. Confirmed by rebinding
`check_correlations` in the namespace and counting: one call where the branch's
own premise implies none. Two consequences, the second reproduced: the
`multinom()` and `gam()` fits still ran, and a predictor of a class those
functions reject — a `Date` column — still aborted the call even though its
correlations had been supplied. `resolve_factor_interactions()` had been given
exactly this treatment in this branch, through its local `factor_correlations()`;
this function had not. The estimate is now computed only when nothing was
supplied. `?generate_model_set` states the replacement semantics.

**With `save.model.fits = FALSE`, a run in which nothing fitted returned a table
of `NA`** rather than the "None of your models fitted successfully" error the
`save.model.fits = TRUE` path raises; `compute_model_weights()` then filled
`delta.AICc` and `wi.AICc` with `NaN`. Which of the two a user got depended only
on a memory setting. The same guard now runs on both paths. This is
long-standing rather than introduced here, but this branch is what restored
`.errorhandling='pass'` on that path, which is what makes total failure
survivable rather than fatal.

**Three validation gaps of the kind this branch set out to close.**
`full_subsets_gam()` validated `r2.type` at entry, on the stated ground that an
unrecognised value should be reported before the candidate set is built, but did
not do the same for `VI.mods`. `progress`, the other argument added in this
release, had no validation at all, so `progress = NA` failed inside a fitting
helper naming neither the argument nor the function. `validate_progress()` sits
in `R/utils.R` with the other shared validators. The `VI.mods` test now asserts
the ordering rather than the rejection, which the previous one did not: it passed
either way, because `fit_model_set()` validates at entry.

**`smooth.smooth.interactions` naming a column of `use.dat` that is not a
predictor** was reported as "not supplied in use.dat", which is not what the
check tests — the branch had moved it from `colnames(use.dat)` to the rows of
the resolved correlation matrix. The message now names the variable and the
requirement. The change of check is itself a fix: such a variable has no row in
the matrix, so `combine_uncorrelated()` took `max()` of an empty sub-matrix and
admitted the `te()` term regardless of its correlations.

`man/FSSgam-package.Rd` had not been regenerated after `DESCRIPTION`'s
Description field was rewritten.

## The suite figures did not reproduce, and are corrected

The branch reported 546 passing and 12 skipped, against a baseline of 438 and
11. Measured at the branch head on Debian (WSL2), R 4.6.1, testthat 3.3.2, with
`testthat::test_local()` and `sum(as.data.frame(res)$passed)`: **568 passing, 0
failing, 7 skipped blocks**, from **447 and 6** on `master`. The differences are
constant at both ends — nine expectations, five blocks — which points at a run
that was regenerating `tests/testthat/_snaps/numerical-regression.md`: creating a
snapshot skips the block, and those five blocks hold nine expectations. Measure
on a clean checkout with the snapshot file present. `NEWS.md`, `CLAUDE.md`, the
two planning documents and the pull request body now carry the same figures,
which is what the Phase 13 note in `CLAUDE.md` asks for.

Line coverage is 94.39% against 94.27% on `master` (`covr`), so the added code
did not dilute it.

## Two other things

The pull request body has been updated and is the current state of the branch.
Beyond the figures above, it now lists `full_subsets_gam()`'s `max.models`
default changing from 500 to 200 under "what changes for a user" — `NEWS.md`
recorded it as a behaviour change but the body did not, and it silently empties
`success.models` for a candidate set of 201 to 500 models.

`planning/RESUME.md` was removed. It was a session handoff note rather than one
of the two planning documents the parent `CLAUDE.md` Section 13 defines, and it
was stale at the branch head in every respect that mattered — "nothing is pushed;
nothing is merged", thirty-three commits against forty-one, and the figures
above. Its one item not recorded elsewhere, that the parallel path must be
checked with a single-cluster script rather than with
`test-fit_model_set_parallel.R`, moved into `CLAUDE.md`, which also gains the
Phase 14 entry for this refactor.

## Not changed

A name in `cyclic.vars` matching no continuous predictor is silently ignored.
Before this branch it could accidentally make a different predictor cyclic,
which is the defect that was fixed; being ignored is the correct outcome for a
name that is not a predictor, but a typo is still not reported. It is recorded
in the Phase 14 entry rather than fixed here, since erroring on it would break
existing code that passes extra names.

https://claude.ai/code/session_014UzX8JZfS6YUUrsLNe6XA4

Issue #14, opened 2026-08-31
parallel = TRUE stalls indefinitely on about half of runs when gamm4 is loaded on the workers


beckyfisher:
## What

`fit_model_set(parallel = TRUE)` stalls indefinitely on roughly half of runs on
this host. The call produces no output and no error; it has to be interrupted.

No code in this package is involved. A `foreach()` whose loop body is `i * 2`
reproduces it, with no FSSgam function called at all. Loading `gamm4` onto the
`doSNOW` workers raises the stall rate from about 1 run in 50 to about 1 in 2.
It is not the sole cause — a stall was observed once without it — but it is by
far the largest factor measured.

## Why it matters

A user who sets `parallel = TRUE` sees the call hang. There is no error to act
on and no indication of which candidate, if any, is at fault, because the
progress bar has not yet drawn anything.

It also affects this repository's own testing. `tests/testthat/test-fit_model_set_parallel.R`
is opt-in behind `skip_unless_parallel_opt_in()` precisely because of this, and
`.github/workflows/parallel-tests.yaml` carries its own `timeout-minutes` so a
stall fails that job alone. That guard is currently the only thing keeping an
unattended stall from consuming the whole runtime of a `covr` or CI job.

## Evidence

Measured on WSL2 (Debian), R 4.6.1, `parallel::makeCluster(2)`, one cluster per
fresh R process, 25 s timeout (a successful run takes 1-2 s). The loop body is
`i * 2` throughout, so no FSSgam code executes.

| `.packages` | completed | loads gamm4 |
|---|---|---|
| `mgcv` | 10/10, 9/10, 20/20 | no |
| `mgcv, MuMIn` | 10/10 | no |
| `gamm4` | 7/10, 5/10 | yes |
| `mgcv, gamm4` | 3/10 | yes |
| `mgcv, FSSgam` | 6/10 | yes, via Imports |
| `mgcv, gamm4, MuMIn, FSSgam` | 8/10 | yes |

Pooled: **1 stall in 50 runs without `gamm4`, 21 in 50 with it.**

Two qualifications. A stall did occur once without `gamm4`, so loading it is not
a precondition, only a strong predictor. And no ordering should be read into the
individual `gamm4` rates — at n = 10 those intervals overlap heavily, and the
rows were not all collected under the same machine load.

The corresponding rate through the package itself, plain script against an
installed copy, `case_study1`, 8 candidate models:

| path | completed |
|---|---|
[...]

PR #18, opened 2026-09-01
Fit the factor-factor null deviance on the pair's complete cases (#16)

### LINKED ISSUE #16: check_correlations compares deviances computed on different row sets when a factor has missing values

beckyfisher:
Closes #16.

## What

`check_correlations()` estimates the correlation between two factor predictors
as `sqrt(1 - fit/null.fit)`. The fitted model was fitted on the pair's complete
cases and the intercept-only model on the whole column, so whenever either
factor carried missing values the two deviances came from different sets of
rows. Both are now fitted on the pair's complete cases.

## Why it matters

A ratio of deviances computed on different numbers of rows does not estimate
anything, so the reported correlation was wrong by an amount that depended on
how much of the column was missing, with no warning. The error is in one
direction: on 40 generated frames with 20 per cent missingness every value fell,
and for independently drawn factors the reported correlation went from about
0.65 to about 0.13.

Values are unchanged for data with no missing predictor values. That is every
call `generate_model_set()` makes, because `check_predictor_missingness()`
rejects predictors containing `NA` first, so the change is confined to direct
calls to `check_correlations()`, which is exported and documented as taking any
data frame.

## Evidence

Against `origin/master` the first new test fails, reporting 0.85 where 0.73 is
expected. Full suite on this branch with `master` merged in: 579 passing, 0
failing, 7 skipped, measured with `testthat::test_local()` and
`sum(as.data.frame(res)$passed)` on Debian WSL2, R 4.6.1, testthat 3.3.2,
against 571 on `master` at fb83312.

The estimate is unchanged where nothing is missing: of 40 generated frames with
no missing values, 40 were `identical()` to `master`.

<details>

### The change

`fill_factor_factor_correlations()` in `R/check-correlations.R`.

Option 1 of the three recorded on the issue, with the cost the issue attributes
to it removed. Fitting the null per pair unconditionally would return the count
of null fits from n to n(n-1). Instead `complete.counts` records how many rows
each factor's own null was fitted on, and the null is refitted only for a pair
whose complete-case count is lower. `na.omit()` of the pair intersects the two
presence sets, so its rows are contained in those on which the response is
present, and a subset of equal size is the same set: equal counts therefore mean
an identical row set and the retained value is exact rather than close.

The refit also requires the pair's own fit to have succeeded. Without that a
failed pair paid for a null whose value is never read, and `try(silent = TRUE)`
suppresses errors but not warnings, so that fit's warnings reached the user.

The hoisted nulls are fitted only for the factors some pair retains. Where every
factor is missing in rows the others are not, no pair retains and fitting all of
them would add n discarded fits to the n(n-1) refits. Measured with
`trace(nnet::multinom)` on three three-level factors over 90 rows: 9 calls with
no missing values, matching `master`; 12 with missingness in disjoint blocks,
which is the same as refitting per pair with no hoist at all.

Option 1 was preferred over option 3 (one `na.omit()` across all factor columns
before the loop) because the rest of the matrix is already pairwise
complete-case: the continuous block comes from
`cor(use = "pairwise.complete.obs")` and the factor-continuous block from a
per-pair `lm()`. Option 3 would have made the factor-factor block the only one
on a different convention, and would have discarded more rows than any single
pair requires.

### Also changed

`null.fit` is now checked for `try-error` alongside `fit`. Previously a failed
null fit with a successful pair fit would have raised an error out of `round()`
rather than leaving the cell `NA`, which is how a failed pair fit is already
handled. The guard is defensive: both models are fitted on the same rows and
`multinom()`'s relevant failure depends only on the response they share, so no
case reaching it is known. That is stated on the face of the code so a later
session does not look for one.

`DESCRIPTION` moves to 1.1.0.9000 to match the development section opened in
`NEWS.md`. `cran-comments.md` records what that does to `R CMD check --as-cran`
and that the version is set to the release number before a submission is built.

`check_non_linear_correlations()` gained the matching `@details` paragraph on
pairwise handling of missing values; its behaviour is unchanged.

### Tests

`tests/testthat/test-check_correlations.R`, three new blocks and one amended.

- The correction itself, asserting the matched value and separately asserting
  that the whole-column value is absent, so a revert fails on both.
- The direction in which the NA-bearing factor is the response, where the pair's
  rows are exactly that factor's own complete cases and the retained value is
  already correct.
- The retained value against the null refitted for every one of six ordered
  pairs over three factors, which is what pins the shortcut. It fails under a
  variant that never refits and under one that indexes `complete.counts` by the
  predictor.
- A column with no observed values, asserting by `multinom()` call count that
  its intercept-only model is not fitted at all, since the returned matrix is
  the same either way.
- The existing `parallel = TRUE` block now gives one factor missing values. The
  refit condition is written once in the `%dopar%` body and once in the
  sequential loop, and `case_study1` has no missing values, so without this the
  parallel copy was never reached. Verified with `FSSGAM_TEST_PARALLEL=true`
  against an installed copy: two runs completed and agreed with the sequential
  result, a third stalled, which is FSSgam_package#14 and not this change.

### Raised separately

- #19, `check_non_linear_correlations()` lacks the same `try-error` guard on its
  intercept-only fit.
- #20, two `DESCRIPTION` URLs redirect and produce a CRAN incoming feasibility
  NOTE. Pre-existing on `master`.

### Not covered

Whether any caller passes factors with missing values in practice. The package's
own functions cannot.

</details>


beckyfisher:
## Review round 1

An independent review of `073ba21` raised two substantive points and one finding
outside the scope of this PR. All are addressed in `b8f4d47`.

## The null was refitted for a pair whose fitted model had already failed

The refit sat before the `try-error` guard, so a pair whose fitted model failed
still ran a `multinom()` whose value is never read. `try(..., silent = TRUE)`
suppresses errors but not warnings, so the discarded fit's warnings reached the
user. Measured on two factors with no overlapping complete cases: 5 warnings on
`073ba21` against 3 on `master`.

The refit condition now requires `fit` to have succeeded as well. The warning
count returns to 3 and no returned value changes, because the cell is `NA`
whenever `fit` fails regardless of what the null holds.

## The second test claimed more than it asserted

The block named "the null model is not refitted when a pair drops no extra rows"
asserts a value and cannot observe whether a refit occurred. It passes against
`master`, where the shortcut does not exist, and against a variant that refits
unconditionally.

It is renamed to what it does assert, and a third block now pins the shortcut by
its result: six ordered pairs over three factors whose missing values fall in
different rows, each compared against the null refitted for that pair. Any
divergence between the retained per-factor value and a refitted one fails there.

## Raised separately

`check_non_linear_correlations()` lacks the `try-error` guard on its own
intercept-only fit, which is the same shape as the guard added here. Raised as
#19 rather than changed in this PR.

## Also changed

`DESCRIPTION` is bumped to 1.1.0.9000, so the development section opened in
`NEWS.md` and the declared version agree. The defensive nature of the `null.fit`
guard is stated on the face of the code, the comment about `multinom()` and
`na.action` is qualified as holding under the default, and
`check_non_linear_correlations()` gained the matching `@details` paragraph on
pairwise handling of missing values.

Suite: 572 passing, 0 failing, 7 skipped, against 568 on `master`.

<details>
<summary>What the review verified, and how</summary>

Reported by the reviewing session, measured on Debian WSL2, R 4.6.1, nnet
7.3-20, doSNOW 1.0.20, testthat 3.3.2, in scratch copies of this branch, of
`origin/master`, and of a variant with the shortcut disabled.

- **The shortcut is exact.** A variant with the refit condition replaced by
  `if(TRUE)` was compared against this branch over 150 generated data frames
  (2-4 factors of 2-4 levels, 40-120 rows, missingness at 0, 5, 20 and 40 per
  cent). All 150 matrices `identical()`. The set argument holds directly as
  well: `na.omit(dat[, c(var.1, var.2)])` intersects the two presence sets, so
  its rows are contained in those on which `var.1` is present, and a subset of
  equal size is the same set.
- **Complete data is unchanged.** The 10 of those 150 frames with no missing
  values were `identical()` to `origin/master`. The other 140 differ, with
  `master`'s values systematically inflated: about 0.65 falling to about 0.13
  for independently drawn factors.
- **`complete.counts` reaches the doSNOW workers.** `parallel = TRUE` on a
  three-factor frame with missing values returned a matrix `identical()` to the
  sequential one, including on the `try-error` paths.
- **No NaN regression.** 270 frames, 120 of them deliberately near-saturated,
  produced no `NaN` cell and no "NaNs produced" warning on either branch.
- **Edge cases checked and unchanged:** a pair whose response has one observed
  level after subsetting, a pair with no overlapping complete cases, and an
  unused predictor level after subsetting. "Equal count, different row set" is
  not constructible.
- `devtools::check()` reports 0 errors, 0 warnings, 0 notes;
  `roxygen2::roxygenise()` leaves the working tree clean.

</details>

beckyfisher:
## Review round 2

An independent review of `b8f4d47` confirmed the round-1 claims and raised four
further points. All are addressed in `d2e865c`. The pull request body has been
rewritten for the current state of the branch, so a review written against the
earlier body is answering different text.

## The hoisted nulls are now fitted only where a pair uses them

Where every factor is missing in rows the others are not, no pair meets the
retain condition, so every hoisted null was discarded. Measured with
`trace(nnet::multinom)` on three three-level factors over 90 rows with
missingness in disjoint blocks: 15 calls, of which 3 were never read.

A factor's null is read only by a pair that drops none of the rows on which that
factor is present, and that is decidable from the missingness pattern before
anything is fitted. It is now decided there. A cache filled on first use would
not have worked, because the `%dopar%` branch runs each pair on a worker that
cannot write back.

The count is now 9 with no missing values, matching `master`, and 12 with
disjoint missingness, which is what refitting per pair with no hoist at all
would cost. The unread fits' warnings go with them: the two-factor frame with no
overlapping complete cases now raises 2 warnings where `master` raises 3, with
the same matrix returned.

## The parallel copy of the changed lines is now reached by a test

The refit condition is written once in the `%dopar%` body and once in the
sequential loop, and the only `parallel = TRUE` fixture was built from
`case_study1`, which has no missing values. Every pair took the retained path, so
the parallel copy was never executed and a later edit to one copy alone would not
have been reported.

`ZONE2` in that fixture is now given missing values, so the two branches are
compared on the refit as well. Verified against an installed copy with
`FSSGAM_TEST_PARALLEL=true`: two runs completed and returned the sequential
result, a third stalled, which is #14 and not this change.

## The new test block's comment misstated its own coverage

The comment claimed that four of the six ordered pairs take the retained value.
On the block's own data two do, `f2` against `f1` and `f3` against `f1`, because
a pair whose only NA-bearing column is the response drops nothing extra. `f2`
with `f3` has 44 complete cases against counts of 65 and 69, so both directions
refit. The comment now gives those counts.

## The version bump is kept, and its consequence recorded

`DESCRIPTION` stays at 1.1.0.9000, which is what the development section of
`NEWS.md` means. `cran-comments.md` was the thing out of date, and now records
that a development version is carried between releases, that this adds one NOTE
under `R CMD check --as-cran` ("Version contains large components"), that
`devtools::check()` does not run that check at all, and that the version is set
to the release number before a submission is built.

## Raised separately

#20, two `DESCRIPTION` URLs redirect and produce a CRAN incoming feasibility
NOTE. Pre-existing on `master`, so not changed here.

Suite: 572 passing, 0 failing, 7 skipped, against 568 on `master`.

beckyfisher:
## Review round 3

An independent review of `d2e865c` found no blocking and no substantive defect.
Every measurement quoted in the round-2 comment reproduced, and the `retains`
precomputation was established exact rather than accepted: an instrumented build
raising an error on a `NULL` null with a successful pair fit was run over 400
generated frames and never fired, recording 200 benign `NULL` reads, all with the
pair fit already failed. A variant with the refit deleted from the `%dopar%` body
alone was built, installed and run against this branch's tests, and reported an
error, which confirms the new fixture does reach the parallel copy.

Four nits were raised and all four are applied in `c62de40`.

## The guard is load-bearing, not defensive

The comment on the `null.fit` guard claimed no case reaching it was known. That
is false. Under `options(na.action = "na.fail")` the pair model is fitted on
`dat.r`, whose incomplete rows have already been removed, while a retained null
is fitted on the raw column and fails. Confirmed directly on two factors, one
carrying `NA`:

| `na.action` | `master` | this branch |
|---|---|---|
| `na.omit` | 0.8477 / 0.6815 | 0.7281 / 0.6815 |
| `na.exclude` | 0.8477 / 0.6815 | 0.7281 / 0.6815 |
| `na.fail` | error, "non-numeric argument to mathematical function" | 0.7281 / `NA` |

The comment now records that, which is a stronger statement about the guard than
the one it replaces.

## Three smaller corrections

`!is.null(null.fit)` is added to both copies of the guard. `inherits(NULL,
"try-error")` is `FALSE`, so the previous form was safe only because `&&`
evaluates the `fit` test first.

The hoisted null is now skipped for a column with no observed values at all. Its
complete-case count is zero, which satisfies the retain equality for every pair,
so it was fitted once for a null that cannot converge and that no pair can read,
every pair it appears in having no rows. Measured on two factors with one
entirely missing: 2 fits on this branch against 3 before and 4 on `master`, with
the same matrix and the same warnings.

`cran-comments.md` said the development version "adds one NOTE". It adds a line
to the incoming feasibility NOTE, which carries "New submission" and the #20 URLs
regardless. It now says so, and attributes the headline 0/0/0 to
`devtools::check()`, which does not run that check at all.

## State

Suite: 572 passing, 0 failing, 7 skipped, against 568 on `master`. Fit counts
unchanged from round 2 on ordinary data: 9 `multinom()` calls with no missing
values, matching `master`, and 12 with disjoint missingness.

beckyfisher:
## Review round 4, and a correction

A fourth independent review, of `c62de40`, found no blocking and no substantive
defect. Rounds 3 and 4 have now both found nothing substantive, so this is the
last round on this pull request.

## Correction to the round-3 comment

That comment said the entirely missing column case returns "the same matrix and
the same warnings". The matrix is `identical()` across `master`, `d2e865c` and
this branch. The warning count is not: `master` and `d2e865c` each raise three
"groups 'a' 'b' 'c' are empty" and this branch raises two, one fewer because one
fewer model is fitted. That is the intended reduction, claimed as a benefit in
round 2; only the round-3 sentence was wrong.

## What the review established

The two clauses added in round 3 cannot change a returned value or raise an
error. `null.deviances[[v]]` is `NULL` in exactly two cases and in both the value
is unreadable rather than merely unread: where no pair retains, the intersection
is strictly smaller than the factor's own count so the refit fires for every pair
in which it is the response; where the count is zero, `nnet::multinom()` on a
zero-row frame fails for both `y ~ x` and `y ~ 1`, so the pair fit has already
failed and the guard short-circuits.

Fuzzed over 400 generated frames including entirely `NA` and all-but-one-observed
columns, against a build with the shortcut deleted and the null refitted per pair
unconditionally: 400 of 400 matrices `identical()`, no error on any build, and no
cell `NA` on this branch where `master` returned a number. The `na.action` table
reproduced exactly, and `na.pass` behaves as `na.fail` does.

## What changed in `19b38bd`

The review recorded that neither new clause was covered: both were deleted in
variant builds and the suite still reported 572 passing.

The zero-count clause is now pinned. It changes a fit count rather than a value,
so it is asserted by counting `multinom()` calls with `trace()`: two pair fits,
both failing on a zero-row frame, and no null at all. Removing the clause makes
that block fail with 3.

`!is.null(null.fit)` is left untested. It is unreachable under the default
`na.action`, so a test would have to set a global option to reach a branch that
exists to convert an error into an `NA`. The reasoning is on the face of the
code.

A misaligned continuation line in the sequential guard was realigned.

Suite: 576 passing, 0 failing, 7 skipped, against 568 on `master`.

beckyfisher:
## Review round 5

A confirmation review of `19b38bd`, the one commit not covered by round 4, found
nothing blocking and nothing substantive.

- The new block is discriminating and is the only one that catches the clause it
  was written for: 2 `multinom()` calls on this branch and 3 with
  `|| complete.counts[[v]] == 0` deleted, with the returned matrix identical on
  both builds, which is what makes the call count the only observable.
- `trace()` and `untrace()` leave no state behind. After the block the `nnet`
  binding is not a `functionWithTrace`, is locked again, and its body is
  identical. The `on.exit()` fires on a pass, on a failing expectation and on an
  error, and `trace()` works with `methods` not attached.
- The realigned continuation line is whitespace only: `git diff -w` is empty and
  the parse trees are `identical()`.
- Suite: 576 passing, 0 failing, 7 skipped, against 568 on `master`.

Rounds 3, 4 and 5 found no blocking defect and rounds 4 and 5 nothing
substantive, so this pull request is finished from my side and is ready for your
review.

beckyfisher:

Five rounds, each by a session with no prior context. Rounds 3, 4 and 5 found no
blocking defect and rounds 4 and 5 nothing substantive. This is the record of
what was found and fixed in the package, not of how it was arrived at.

## Defects corrected in this pull request

**The reported correlation between two factors was wrong whenever either carried
missing values.** `check_correlations()` estimates it as `sqrt(1 - fit/null.fit)`,
with the fitted model on the pair's complete cases and the intercept-only model
on the whole column, so the ratio was taken between deviances computed on
different numbers of rows. On 40 generated frames at 20 per cent missingness every
off-diagonal value fell, from about 0.65 to about 0.13 for independently drawn
factors. Both models are now fitted on the pair's complete cases. Values are
unchanged where nothing is missing, which is every call `generate_model_set()`
makes.

**A pair whose fitted model failed still paid for the null.** The refit was
placed before the `try-error` guard, and `try(silent = TRUE)` suppresses errors
but not warnings, so a discarded fit's warnings reached the user. Found in
round 1: 5 warnings against `master`'s 3 on a pair with no overlapping complete
cases.

**A failed null fit could reach `round()` as a character vector.** Guarded, and
the guard turns out to be load-bearing rather than defensive: under
`options(na.action = "na.fail")` it converts `master`'s "non-numeric argument to
mathematical function" into an `NA` cell. Found in round 3.

**A column with no observed values was fitted once for a null nothing can read.**
Its complete-case count is zero, which satisfies the retain condition for every
pair. Found in round 3; the case now costs 2 `multinom()` calls against
`master`'s 4.

## What the reviews established rather than corrected

The shortcut that keeps the hoisted per-factor null for a pair that drops no
extra rows is exact, not approximate: 400 generated frames including entirely
`NA` and all-but-one-observed columns are `identical()` to a build that refits
the null for every pair. The parallel and sequential copies agree, including
where `null.deviances` crosses to the workers carrying `NULL` elements.

## Raised separately

#19, `check_non_linear_correlations()` lacks the same guard on its intercept-only
fit.

## State

576 passing, 0 failing, 7 skipped, against 568 on `master`. `devtools::check()`
0/0/0.

beckyfisher:
This looks ok.

beckyfisher:
`master` was merged into this branch to resolve the conflicts. Both conflicts
were in files that this branch and the #14 work both append to, and both sides
were kept:

- `NEWS.md`, the development section. The `gamm4`/`Suggests` and worker-packages
  entries from #14 come first, then this branch's `check_correlations()` entry.
- `prompts/post-refactor-issues-14-15-16.md`, the shared session log. The
  opening response listing the three issues and the three decisions is kept in
  the fuller of its two forms; the #14 and #16 narratives follow in that order.
  The #16 review-round headings are now qualified by issue number, since the
  file records two issues' rounds.

`DESCRIPTION` and `cran-comments.md` merged without conflict: both branches made
the same development-version change.

The Evidence section of the body was updated, since its figures were measured
before the merge. Remeasured on the merge commit, same host and versions: 579
passing, 0 failing, 7 skipped, against 571 on `master` at fb83312. Nothing else
in the body moved.

Issue #20, opened 2026-09-01
DESCRIPTION URLs redirect, producing a CRAN incoming feasibility NOTE


beckyfisher:
## What

Two of the three URLs in `DESCRIPTION` are written without a trailing slash and
redirect. `R CMD check --as-cran` reports them under checking CRAN incoming
feasibility.

## Why it matters

CRAN submission is the stated end goal for this package. The incoming
feasibility check runs on submission, so this NOTE would have to be corrected or
explained then. Correcting it is a two-character change to `DESCRIPTION` and a
re-run of roxygen2 for `man/FSSgam-package.Rd`.

## Evidence

`R CMD check --no-manual --as-cran` on a tarball built from `master`, Debian
WSL2, R 4.6.1, with network access:

```
Found the following (possibly) invalid URLs:
  URL: https://beckyfisher.github.io/FSSgam (moved to https://beckyfisher.github.io/FSSgam/)
    From: DESCRIPTION
          man/FSSgam-package.Rd
    Status: 301
  URL: https://beckyfisher.github.io/FSSgam_package (moved to https://beckyfisher.github.io/FSSgam_package/)
    From: DESCRIPTION
          man/FSSgam-package.Rd
    Status: 301
```

The NOTE does not appear with `_R_CHECK_CRAN_INCOMING_REMOTE_=false`, and does
not appear under `devtools::check()`, which disables the incoming checks
entirely, so it is not visible in the workflow this repository normally uses.

<details>
<summary>Implementation detail</summary>

Add the trailing slash to both `https://beckyfisher.github.io/FSSgam_package`
and `https://beckyfisher.github.io/FSSgam` in `DESCRIPTION`'s `URL` field, then
run `devtools::document()` so `man/FSSgam-package.Rd` follows.
`https://github.com/beckyfisher/FSSgam_package` is not redirected and needs no
change.

Verify with `R CMD build .` followed by
`R CMD check --no-manual --as-cran FSSgam_*.tar.gz` with network access; the URL
half of the incoming feasibility NOTE should be gone.

</details>

PR #21, opened 2026-09-01
Compute cor.matrix rows for the interaction columns a user cannot name (#15)

### LINKED ISSUE #15: A supplied cor.matrix must name factor interaction columns the user cannot know in advance

beckyfisher:
Closes #15.

## What

With `factor.factor.interactions` set, a user-supplied `cor.matrix` had to carry
a row and a column for every hard coded interaction column the setting causes to
be created, or `generate_model_set()` stopped. Rows and columns for any of those
the matrix does not carry are now computed from `use.dat` and appended.

## Why it matters

The names are deterministic, but which of them exist is not. An interaction
column is built only for a combination of factors that survives the collinearity
screen, and that screen reads the supplied matrix, so a user cannot construct
the matrix without already knowing what the matrix will produce. The failure
arrived partway through model set construction and named columns that do not
exist in the user's own data.

A predictor the user named in `pred.vars.cont`, `pred.vars.fact` or
`linear.vars` and did not supply is still an error, so a misspelled name is
still reported rather than quietly computed.

## Evidence

The issue's reproduction, three factors reduced to two for brevity: `ZONE` and
`ZONE.copy` as factors with a supplied 3x3 matrix over `depth`, `ZONE` and
`ZONE.copy`. On `master` this stops with

```
Error in build_predictor_correlation_matrix(...) :
  Supplied cor.matrix is missing required predictors: ZONE.I.ZONE.copy.
```

On this branch it returns a model set whose `predictor.correlations` is the
supplied 3x3 unchanged, with a fourth row and column for `ZONE.I.ZONE.copy`
holding 1.000 against each of its components and 0.082 against `depth`, which is
what `check_correlations()` gives for that column.

Twelve of the fifteen new test blocks fail against `origin/master`, eight of them with that error.
Full suite on this branch with `master` merged in: 616 passing, 0 failing, 7
skipped, measured with `testthat::test_local()` and
`sum(as.data.frame(res)$passed)` on Debian WSL2, R 4.6.1, testthat 3.3.2,
against 579 on `master` at 94bef22.

<details>

### The change

`R/generate-model-set.R`.

`resolve_factor_interactions()` now returns `factor.interaction.terms`, the
names of the columns it appended to `use.dat`. That is what lets
`build_predictor_correlation_matrix()` tell a name the user could have supplied
from one the function invented; the alternative, matching on the `.I.`
separator, would also match a user's own column that happened to be named that
way.

`build_predictor_correlation_matrix()` gains a `derived.predictors` argument and
splits the missing predictors on it. Named ones stop the call, with the same
message as before. Derived ones go to `augment_supplied_correlation_matrix()`,
which computes a matrix over `all.predictors` and splices in only the rows and
columns involving a derived column, so every cell the user supplied is kept as
supplied.

Option 2 of the three recorded on the issue. Option 1, falling back for any
missing predictor, was rejected because it would compute a value for a predictor
the user misspelled instead of reporting it, which is what the existing check is
for. Option 3, documenting the construction, leaves the error arriving partway
through construction and still requires the user to anticipate the screen.

### The one thing this cannot do

Computing a derived column's correlations reads every other predictor's data, so
the FSSgam_package#13 allowance for a predictor of a class `check_correlations()`
rejects does not extend to this case. The error says that, naming the interaction
columns and carrying the underlying class error, rather than reporting the class
error alone. A user in that position supplies the interaction columns as well,
which is what the documented contract required of everyone until now.

A name the supplied matrix carries that is not a predictor of this model set has
no computed value against a derived column. Those cells are set to zero, which is
what the branch that computes the whole matrix already does with the `NA` values
`check_correlations()` returns. Leaving them `NA` is not safe: the character form
of `smooth.smooth.interactions` validates its argument against the rownames of
the resolved matrix rather than against the predictors, so such a name can reach
`combine_uncorrelated()`, where `max()` of an all-`NA` sub-matrix returns `NA` and
`if (NA)` raises an error naming neither the matrix nor the argument. An `NA` the
user supplied among their own predictors is left alone.

The supplied names are kept in their own order per dimension, with the derived
ones appended to each, so a matrix whose rownames and colnames differ is not
reshaped into a square one carrying an invented all-`NA` row. Which names are
missing is also tracked per dimension, and each splice is indexed with its own
set: a derived name the user supplied as a column and not as a row keeps the
values they gave for that column, and only the row is computed.

A duplicated name in a supplied matrix is reported rather than resolved
arbitrarily, for any name the model set indexes. Which of two rows of the same
name governs a screen depends on how the sub-matrix is indexed, and the
augmentation would keep only one of them. A duplicate among names nothing reads
is accepted, as it is when no matrix is supplied.

### One pre-existing defect fixed here

`combine_uncorrelated()` and `enumerate_candidate_models()` built the correlation
sub-matrix for a candidate from two positional indices, one in rowname order and
one in colname order. Where a matrix carries the same names in different orders in
its two dimensions, the diagonal of that sub-matrix holds cross correlations and
the triangles hold the ones from the original diagonal, so the candidate is
dropped whatever its predictors' correlation. On `master`, a supplied matrix over
`depth`, `complexity` and `ZONE` with its columns permuted returns 5 candidates
where the same matrix unpermuted returns 7.

It is fixed here rather than deferred because the partial supply this change
accepts is what produces the divergence from a matrix the user wrote
consistently: the derived names are appended to the end of each dimension, so a
derived name supplied as a column anywhere other than last leaves the two orders
different. Both sites now index by name, both dimensions in one order. The golden
master is 31 of 31 identical across the change, every scenario there supplying
its dimnames in one order. The right hand side
of the splice is coerced with `as.matrix()`, because assigning a `data.frame`
into a sub-block of a matrix replaces the matrix with a plain list.

### Tests

`tests/testthat/test-generate_model_set_interactions.R`, fifteen blocks. Twelve
fail against `origin/master`'s `R/`. Three were written for the issue itself:

- the issue's reproduction, asserting that the supplied block is unchanged and
  that the derived row and column match `check_correlations()` on the same data;
- the same under `non.linear.correlations = TRUE`;
- the combined case of an unclassifiable predictor and a missing interaction
  column, asserting the error names the reason.

Nine cover the review rounds: a `data.frame` `cor.matrix`, a supplied name that is
not a predictor, a matrix whose two dimnames differ, a derived name supplied in
one dimension only, a duplicated name that is a predictor and one that is not,
the position of a supplied derived name within its dimension, a complete matrix
with its columns permuted, and a continuous predictor omitted alongside a missing
derived column. All but the last fail against the commit that preceded them on
this branch; the last fails under a mutation folding named-missing into
derived-missing, which is what it exists to catch.

Two pin behaviour the fix must not take away and pass against the pre-fix code
as well, which is recorded on the face of each so neither is mistaken for a
regression test: that a supplied value for an interaction column is used rather
than recomputed, asserted by supplying zeros for a pasted column the data say is
perfectly collinear with its components and checking that a candidate holding
both survives; and that a predictor the user named and omitted is still reported
by name, with the interaction column absent from the same message.

### Also changed

`DESCRIPTION` moves to 1.1.0.9000 to match the development section of `NEWS.md`,
and `cran-comments.md` records what that does to `R CMD check --as-cran`. Both
are identical to the corresponding change on #18, so the two merge without
conflict in either order.

</details>


beckyfisher:
## Review round 1

An independent review of `512d79c` found one blocking defect and three
substantive points. All are addressed in `4720f7f`. The pull request body has
been rewritten for the current state.

## A cor.matrix supplied as a data.frame crashed

Assigning a `data.frame` into a sub-block of a matrix does not coerce the
`data.frame`; it replaces the matrix with a plain list, dropping `dim` and
`dimnames`, and the next assignment then fails with "incorrect number of
subscripts on matrix", naming neither the argument nor the cause. A `data.frame`
works on `master`, and worked on this branch whenever the supplied matrix was
complete, so the one class of input the feature did not work for was the one it
was added for.

The right hand side of the splice is coerced with `as.matrix()`.

## The NA cells were reachable

The comment claimed those cells are never read, on the ground that every stage
that screens on correlation indexes the matrix by the terms of a candidate model.
That is right for `enumerate_candidate_models()` and for the logical form of
`smooth.smooth.interactions`, and wrong for the character form, which validates
its argument against the rownames of the resolved matrix rather than against the
predictors. A supplied name that is not a predictor therefore reaches
`combine_uncorrelated()`, where `max()` of an all-`NA` sub-matrix returns `NA` and
`if (NA)` raises "missing value where TRUE/FALSE needed". `master` stops earlier
and cleanly, so the failure mode was new.

The cells the augmentation itself creates are now set to zero, which is what the
branch that computes the whole matrix already does with the `NA` values
`check_correlations()` returns. An `NA` the user supplied among their own
predictors is left alone. The two branches now leave the resolved matrix with the
same guarantee.

The related case the same change covers: a supplied matrix whose rownames and
colnames differ was reshaped into a square one carrying an invented all-`NA` row,
which was returned to the user as `predictor.correlations`. The supplied names
are now kept in their own order per dimension, with the derived ones appended to
each.

## The named/derived split had no discriminating test

The split is the guarantee the body leads with, that a predictor the user named
and omitted is still reported rather than quietly computed. The block written for
it observes an error raised by `factor_correlations()` several stages earlier,
because it omits a factor, so it passed whatever the split did. A mutation
folding named-missing into derived-missing passed the entire suite.

The case that reaches `build_predictor_correlation_matrix()` with both kinds
missing is a continuous predictor omitted alongside a derived column. A block now
covers it, asserting that the message names `complexity` and not
`ZONE.I.ZONE.copy`. It fails under that mutation.

## The parameter documentation contradicted itself

`@param cor.matrix` stated the new exception and then retained, unqualified, the
#13 sentence saying `check_correlations` is not called at all. It is now
qualified rather than deleted, since it remains true for a complete matrix.

## Raised separately

- #22, a generated interaction column name can collide with an existing column of
  the same name, and `cbind()` appends a duplicate rather than reporting it. This
  is the residual case behind the review's observation that the named/derived
  split is by name. Pre-existing on `master`.
- The two redirecting `DESCRIPTION` URLs the review found in `cran-comments.md`
  are #20, already open. That file is byte-identical to #18's copy and is left
  alone here so the two merge without conflict.

`NEWS.md` does conflict with #18, both branches inserting a bullet after a new
development heading. The resolution is to keep both bullets.

Suite: 594 passing, 0 failing, 7 skipped, against 568 on `master`.

beckyfisher:
## Review round 2

An independent review of `4720f7f` confirmed all four behavioural claims from
round 1 and found one substantive defect remaining. It is fixed in `b1ca7d3`,
along with one nit worth taking. Two figures quoted in the round 1 comment and in
the body were wrong and are corrected below.

## A derived name supplied in one dimension only was overwritten

The set of missing names was pooled over the two dimensions, so a derived name
the user supplied as a column and not as a row counted as missing from both, and
the splice replaced the column they had supplied with computed values. That is
the guarantee this change is justified by, failing in exactly the case round 1
added support for.

The missing names are now tracked per dimension and each splice is indexed with
its own set. A test supplies `ZONE.I.ZONE.copy` as a column only, with the values
0.11, 0.22 and 0.33, and asserts they survive; it fails against `4720f7f`.

One correction to the review's account of the consequence. The values in
`predictor.correlations` were overwritten, but the candidate set was not changed
in that configuration. `enumerate_candidate_models()` screens both triangles, and
the row was not supplied, so the computed 1 excludes `ZONE+ZONE.I.ZONE.copy`
whichever way the column is resolved. The test records that rather than asserting
a difference that does not occur.

## A duplicated name is now reported

A supplied matrix carrying the same name twice was reduced by `unique()` during
augmentation, keeping one of the two silently. Which of them governs a screen
depends on how the sub-matrix is indexed, so the matrix is ambiguous rather than
merely redundant, and it is now reported by name.

## Two figures that did not reproduce

Both were quoted rather than measured, which `CLAUDE.md` section 6 warns about.

The round 1 comment said a mutation folding named-missing into derived-missing
passed the entire suite. Applied literally it fails a pre-existing block,
`generate_model_set errors when a supplied cor.matrix omits a predictor`. The
narrower mutation the review supplies, conditioned on `derived.predictors` being
non-empty, does pass at `512d79c` and is caught by the new block, so the reason
for adding that block stands and the sentence describing the evidence did not.

The body said three of the new blocks fail against `master`. Re-measured by
running the branch's interaction test file against `master`'s `R/`: nine of the
eleven fail, six of them with the missing-predictors error. The body now says so.

## Raised separately

- #26, a 1x1 supplied `cor.matrix` is silently discarded and recomputed, because
  `length(cor.matrix)==1` is the sentinel for the `NA` default.
- #27, an `NA` between two predictors in a supplied matrix fails with "missing
  value where TRUE/FALSE needed", naming neither the matrix nor the pair, and
  only when `max.predictors` is large enough for that pair to be enumerated.

Both reproduce unchanged on `master`.

## Nits not taken, and why

The augmentation computes every pair to obtain a few rows. Restricting it would
mean reimplementing pairwise estimation rather than calling
`check_correlations()`, since the correlation between a derived column and each
other predictor needs those predictors' data either way. The waste is confined to
the pairs the user supplied, and only when a derived column is missing.

An `NA` dimname now stops with `subscript out of bounds` where `master` tolerated
it, and `predictor.correlations` is returned as a matrix in the augmented path
where a supplied `data.frame` is returned unchanged otherwise. Both are matters
of an ill-formed or unusual input; neither is a wrong answer.

Suite: 599 passing, 0 failing, 7 skipped, against 568 on `master`.

beckyfisher:
## Review round 3

An independent review of `b1ca7d3` verified every claim, including the round 2
correction about the candidate set, which the round 1 review had asserted the
opposite of. Two substantive findings remained; both are fixed in `17c1973`.

## The screens indexed the correlation sub-matrix by position

`combine_uncorrelated()` and `enumerate_candidate_models()` built a candidate's
correlation sub-matrix from two positional indices, one in rowname order and one
in colname order. Where a matrix carries the same names in different orders in
its two dimensions, the diagonal of that sub-matrix holds cross correlations and
the triangles hold the ones from the original diagonal, so the candidate is
dropped whatever its predictors' correlation.

The defect is pre-existing. On `master`, a supplied matrix over `depth`,
`complexity` and `ZONE` with its columns permuted returns 5 candidates where the
same matrix unpermuted returns 7.

It is fixed here rather than deferred because this change is what makes it
reachable from a matrix the user wrote consistently. The derived names are
appended to the end of each dimension, so supplying a derived name as a column
anywhere other than last leaves the two orders different: the review measured 18
candidates with the derived column last and 16 with it first, from matrices with
identical numeric content.

Both sites now index by name, both dimensions in one order. The golden master is
31 of 31 identical across the change, every scenario there supplying its dimnames
in one order, which is what makes the fix safe to make in this pull request.

## The duplicated-name check was broader than its justification

Only a duplicated name the model set indexes is ambiguous. The check stopped the
call for any duplicate, including one nothing reads, which `master` accepts. It
is now restricted to `all.predictors`.

## Tests

Three blocks added, all three failing against `b1ca7d3`: that the position of a
supplied derived name within its dimension does not change the model set; that a
complete supplied matrix screens the same however its columns are ordered; and
that a duplicated name that is not a predictor is accepted.

## Corrections

The round 2 comment said six of the nine blocks failing against `master` carry
the missing-predictors error. Seven do. The body now records fourteen new blocks,
nine of them failing against `master`.

The nit that the duplicate check runs after `resolve_factor_interactions()` has
already indexed the matrix is left as it is. The call still stops before
returning anything, so no wrong answer results; moving the check earlier would
mean validating the matrix before the predictors it is validated against are
known.

Suite: 602 passing, 0 failing, 7 skipped, against 568 on `master`.

beckyfisher:
## Review round 4

An independent review of `17c1973` found nothing blocking and nothing
substantive. Rounds 3 and 4 have now both found no blocking defect and round 4
none substantive, so this is the last round on this pull request. Four nits were
raised and all four are addressed in `e987dee`.

## The name-based indexing was attacked directly and holds

The review established that membership is identical between the old positional
form and the new one, that the ordering difference is a symmetric permutation
wherever the two dimnames agree, and that both triangles are screened, so the
outcome is unchanged in every case where the orders already coincide. That is why
the golden master is 31 of 31 identical across the change. Non-square selection is
unreachable at both sites, and the constructed non-square cases give
byte-identical sub-matrices under both forms.

## One behaviour change was undocumented

A factor named twice in `pred.vars.fact` used to select a 1x1 sub-matrix whose
triangles are empty, so `max()` warned "no non-missing arguments to max" and
returned `-Inf`, the combination survived, and a `ZONE.I.ZONE` column was built.
Indexed by name it is the 2x2 whose off-diagonal is the factor's correlation with
itself, which is 1, so the combination is dropped as perfectly collinear. That is
now tested and recorded in `NEWS.md`.

The other half of that input is unchanged and is not fixed here.
`enumerate_candidate_models()` deduplicates a candidate's terms before indexing,
so it still reaches a 1x1 sub-matrix and still admits a nonsense `ZONE+ZONE`
candidate with two `-Inf` warnings. Raised as #28, with the observation that
rejecting a duplicated entry in `pred.vars.cont`, `pred.vars.fact` or
`linear.vars` up front is a clearer error than anything downstream can give.

## Two figures re-measured

Ten of the fifteen new blocks fail against `master`'s `R/`, eight of them with
the missing-predictors error, not nine and seven. The permuted-matrix candidate
counts in `NEWS.md` now state the `max.predictors` they were measured at: 7
against 5 holds at 2, and 8 against 5 at the default of 3.

## The remaining nit, not taken

The augmentation's `unique()` over the dimnames drops a duplicated name that is
not a predictor from the returned `predictor.correlations`, where the
unaugmented path keeps it. Nothing reads that name, and reporting it would
reinstate the over-broad error round 3 removed.

Suite: 604 passing, 0 failing, 7 skipped, against 568 on `master`.

beckyfisher:
## Review round 5

A confirmation review of `e987dee` found nothing blocking. The new block is
discriminating, the golden master moves in exactly one scenario and that move is
intended, and the suite figures reproduce. One substantive point and one figure
remained; both are corrected in `abd4613`.

## The candidate counts still did not reproduce

The figures round 4 was written to correct were themselves conditional. The
permuted count of 5 reproduces, but the unpermuted 7 and 8 hold only with
`factor.smooth.interactions` switched off. `generate_model_set()` defaults it to
`pred.vars.fact`, so a reader following the entry as written, with everything at
its default, gets 9 and 13.

`NEWS.md` now names the arguments the measurement was taken under and quotes 5
against 9 at `max.predictors = 2` and 5 against 13 at 3, with this branch
agreeing at 9 and 13 in both orders. The qualitative claim was never in doubt:
the permuted matrix loses candidates on `master` at every setting tried, and the
two orders agree on this branch at every setting.

## Twelve blocks fail against master, not ten

The earlier count was read off testthat's summary reporter, which lists at most
ten problems by default. Counted from the result object, twelve of the fifteen
new blocks fail against `master`'s `R/`, producing thirteen reported problems
because one block fails both its expectations. The count of eight carrying the
missing-predictors error does reproduce.

That is the third figure on this pull request that was quoted rather than
measured. The reporter cap is worth recording as the cause of this one.

## Also changed

Both assertions in the duplicated-factor block were negative, so a degenerate
return would have satisfied them. A positive assertion was added alongside.

Suite: 605 passing, 0 failing, 7 skipped, against 568 on `master`.

Rounds 3, 4 and 5 found no blocking defect, and rounds 4 and 5 nothing beyond
figures and comments, so this is the last round on this pull request.

beckyfisher:

Five rounds, each by a session with no prior context. Rounds 4 and 5 found no
blocking defect and nothing substantive beyond figures. This is the record of
what was found and fixed in the package.

## The issue itself

A supplied `cor.matrix` had to name the hard coded `a.I.b` interaction columns
that `factor.factor.interactions` creates, but which of those exist depends on
the supplied matrix, through the screen that decides which factor combinations
survive. Rows and columns for any the matrix does not carry are now computed from
`use.dat` and appended. A predictor the user named and omitted is still reported,
so a misspelled name is not quietly computed.

## Defects found during review and corrected here

**A `cor.matrix` supplied as a `data.frame` crashed.** Assigning a `data.frame`
into a sub-block of a matrix replaces the matrix with a plain list, dropping
`dim` and `dimnames`, so the next assignment failed with "incorrect number of
subscripts on matrix". A `data.frame` works on `master`. Round 1.

**A supplied name that is not a predictor could reach `combine_uncorrelated()` as
`NA`.** The character form of `smooth.smooth.interactions` validates against the
rownames of the resolved matrix rather than against the predictors, so `max()` of
an all-`NA` sub-matrix returned `NA` and `if (NA)` raised "missing value where
TRUE/FALSE needed", naming nothing. The cells the augmentation creates are now
zeroed, as the branch that computes the whole matrix already does. Round 1.

**A matrix whose two dimnames differ was reshaped into a square one** carrying an
invented all-`NA` row, returned to the user as `predictor.correlations`. Round 1.

**A derived name supplied in one dimension only had those values overwritten.**
The missing names were pooled across the two dimensions. They are now tracked per
dimension. Round 2.

**The two collinearity screens indexed the sub-matrix by position.**
`combine_uncorrelated()` and `enumerate_candidate_models()` built it from two
positional indices, one in rowname order and one in colname order, so a matrix
carrying the same names in different orders in its two dimensions had cross
correlations on the sub-matrix diagonal and the original diagonal's ones in the
triangles, and the candidate was dropped whatever its correlation. On `master`, a
supplied matrix with its columns permuted returns 5 candidates where the same
matrix unpermuted returns 9. Pre-existing, and fixed here because the partial
supply this change accepts is what makes it reachable from a matrix the user
wrote consistently. The golden master is 31 of 31 identical across the fix.
Round 3.

A consequence of that fix, tested and recorded: a factor named twice in
`pred.vars.fact` no longer yields an interaction of that factor with itself.

**A duplicated name in a supplied matrix was collapsed silently** by the
augmentation, keeping one of the two. It is now reported, for any name the model
set indexes. Round 2, narrowed in round 3 after it was found to reject input that
works on `master`.

## Raised separately

#22, a generated interaction column name can collide with an existing column and
`cbind()` appends a duplicate. #26, a 1x1 supplied `cor.matrix` is silently
discarded through the `length(cor.matrix) == 1` sentinel. #27, an `NA` between
two predictors in a supplied matrix fails with a message naming nothing. #28, a
predictor named twice yields a candidate holding it twice. All four reproduce
unchanged on `master`.

## State

605 passing, 0 failing, 7 skipped, against 568 on `master`. Twelve of the fifteen
new blocks fail against `master`'s `R/`. Golden master 30 of 31 identical, the one
move being the intended error-message change.

beckyfisher:
This looks ok.

beckyfisher:
`master` was merged into this branch to resolve the conflicts, after #18 landed,
so this now sits on top of both #29 and #18. Both conflicts were in the two files
every branch in this series appends to, and both sides were kept in each:

- `NEWS.md`, the development section. The `gamm4`/`Suggests` and worker-packages
  entries from #14 and the `check_correlations()` entry from #16 come first,
  then this branch's two entries.
- `prompts/post-refactor-issues-14-15-16.md`, the shared session log. The #15
  narrative is appended after the #14 and #16 ones already on `master`, and
  `master`'s fuller form of the shared opening response is the one retained.

`DESCRIPTION` and `cran-comments.md` merged without conflict, as the body
predicted.

The Evidence section of the body was updated, since its figures were measured
before the merge and against a `master` that predates #18. Remeasured on the
merge commit, same host and versions: 616 passing, 0 failing, 7 skipped, against
579 on `master` at 94bef22. Nothing else in the body moved.

PR #29, opened 2026-09-01
Take gamm4 off the parallel workers, and out of Imports (#14)

beckyfisher:
Addresses #14, and does not close it.

## What

`gamm4` moves from `Imports` to `Suggests`, and is no longer loaded onto the
`doSNOW` workers by `fit_model_set(parallel = TRUE)`.

## Why it matters

Two reasons, and the first stands on its own.

Nothing in this package calls `gamm4`. The only uses are `inherits(x, "gamm4")`
class tests, which need no package, and an `@importFrom gamm4 gamm4` directive
for a function never called. That directive made every parallel worker load
`gamm4` and `lme4` through the `FSSgam` namespace, for every model set whatever
the family, and the `Imports` entry made every user install them.

This is a user-visible dependency change. A fresh install no longer brings
`gamm4` or `lme4`, so anyone fitting through `MuMIn::uGamm(lme4 = TRUE)`, as the
companion repository's case study 2 does, has to install `gamm4` themselves.
`NEWS.md` says so in the first sentence.

Loading `gamm4` onto the workers is also the only direction recorded on #14 that
measured as helping the dispatch stall, in which the call hangs indefinitely with
no output and no error. The stall is reduced, not removed, and #14 stays open.

A `uGamm(lme4 = TRUE)` refit still works on a worker, and the measured reduction
does not extend to it. Unserialising such a test.fit loads `gamm4` and `lme4` on
the worker by itself, and `doSNOW` sends the export environment in the same call
that carries the package vector, before the packages are attached, so for those
users the worker side is substantially unchanged. What the change removes is
every other model set paying for it.

## Evidence

`fit_model_set(parallel = TRUE)` on `case_study1`, eight candidates, run from
installed copies of each version alternating in blocks of ten so that both arms
saw the same machine load, one cluster per fresh R process, which is the
condition #14 records as necessary to reproduce the stall at all:

| version | stalled |
|---|---|
| `master` | 44 of 140 |
| this branch | 25 of 140 |

Chi-squared with continuity correction, p = 0.013. Debian WSL2, R 4.6.1,
2026-09-01. A successful run takes 2 to 6 seconds.

The narrower comparison agrees. A `foreach()` whose body is `i * 2`, so that no
package code runs at all, run 50 times with `gamm4` in the packages the workers
attach and 50 times without, alternating run by run:

| workers attach | stalled |
|---|---|
| `mgcv`, `gamm4`, `MuMIn`, `FSSgam` | 34 of 50 |
| `mgcv`, `MuMIn`, `FSSgam` | 19 of 50 |

Chi-squared p = 0.005.

Both are one session's measurement of a rate on one host and should be read as
such. An earlier version of this description quoted the narrower comparison run
with its two arms one after the other rather than alternated; an independent
replication could not reproduce it, which is what that design predicts, and only
alternating designs are quoted now.

That `library(FSSgam)` no longer loads `gamm4` or `lme4` is not a rate and does
reproduce on demand: it is `FALSE` for both from this branch's installed copy and
`TRUE` for both from `master`'s.

<details>

### Why the two halves are one change

Loading the `FSSgam` namespace on a worker loads its imports. A worker given
`.packages = c("mgcv", "MuMIn", "FSSgam")` therefore still loaded `gamm4` while
`gamm4` was under `Imports`, so narrowing the vector alone would have achieved
nothing. #14's own table records this: `mgcv, FSSgam` completed 6 of 10, against
39 of 40 for `mgcv` alone.

`MuMIn` also only suggests `gamm4`, so nothing else pulls it in.

### `worker_packages()`

`R/utils.R`, shared the way `classify_correlation_predictors()` and
`resolve_candidate_family()` are. It takes no argument and returns
`c("mgcv", "MuMIn", "FSSgam")`.

`mgcv` is necessary. `fit_mod_l()` refits through `stats::update()`, which for a
`gam()` test.fit evaluates the stored call by lexical lookup ending at the
worker's search path, and the symbol in that call is `gam`. Measured: 0 of 4
candidates refit without it, failing with `could not find function "gam"`.

`MuMIn` is not, and is kept defensively. A `uGamm` test.fit dispatches to MuMIn's
own `update.gamm` method, whose body runs inside that namespace where `uGamm` is
bound, and that namespace is loaded by unserialising the test.fit. Measured: 8 of
8 candidates refit with `MuMIn` absent from the vector. It stays because a
`test.fit` whose class carries no such method would need it, and attaching an
already-loaded namespace costs nothing.

`gamm4` is absent even for a `test.fit` that is one. An earlier version of this
branch added it back in that case, on the reasoning that such a refit needs it.
It does not: `MuMIn::uGamm()` calls `require("gamm4")` itself, and in any case
unserialising the test.fit has already loaded it.

### The other directions #14 records

Measured first, `foreach()` body `i * 2`, one cluster per fresh R process:

| variant | stalled |
|---|---|
| baseline | 17 of 70 |
| `makePSOCKcluster(setup_strategy = "sequential")` | 9 of 70 |
| `clusterCall()` preload, no `.packages` | 10 of 70 |
| both together | 7 of 50 |

None separates from the baseline at that number of replicates. These arms were
run one after the other rather than alternated, so they carry the weakness
described above and are reported only as the reason those directions were not
pursued further. `setup_strategy` also postdates R 4.0 while `DESCRIPTION`
declares `R (>= 3.5)`, a floor #31 records as unreachable in any case.

### Tests

`tests/testthat/test-utils.R`, two blocks: that `worker_packages()` carries
`mgcv`, `MuMIn` and `FSSgam` and not `gamm4`, and that `gamm4` appears in none of
`Depends`, `Imports` or `LinkingTo`. The second is the invariant the first
depends on, and it fails if anyone makes `gamm4` a hard dependency again;
verified by installing a doctored source with `gamm4` under `Depends`. The other
half of that invariant, re-adding the `@importFrom` while `gamm4` stays under
`Suggests`, is covered by `R CMD check` rather than here.

`skip_if_not_installed("gamm4")` was added to `fixture_coral_ugamm()` and to the
one direct `uGamm(lme4 = TRUE)` call, without which `R CMD check` fails when the
suggested package is absent, which CRAN requires it not to. Verified against a
library view built by symlinking every package in the site library except
`gamm4`: 547 passing, 0 failing, 0 errors, 14 skipped, against 571 and 7 with
`gamm4` present.

The stall itself has no test. It is a property of the host and the cluster, it
takes a rate across many fresh processes to measure, and the suite's own
`parallel = TRUE` blocks are already opt-in behind
`skip_unless_parallel_opt_in()` for that reason.

### Also changed

`DESCRIPTION` moves to 1.1.0.9000 and `cran-comments.md` records what that does
to `R CMD check --as-cran`. Both are byte-identical to the corresponding change
on #18 and #21, so the three merge without conflict in those two files in any
order. `NEWS.md` conflicts with both, each branch inserting a bullet after a new
development heading; the resolution is to keep all the bullets.

### Raised separately

#30, a `dsm` test.fit cannot be refitted under `parallel = TRUE`, because `dsm`
is never loaded onto the workers. Pre-existing, and `worker_packages()` is the
natural place to fix it.

#31, `Depends: R (>= 3.5)` is not reachable, since `MuMIn` requires
`R (>= 4.4.0)`. That weakens the reason given above for rejecting
`setup_strategy = "sequential"`.

#20, two `DESCRIPTION` URLs redirect and are reported by `--as-cran`.

### An unclaimed beneficiary

`check_correlations(parallel = TRUE)` dispatches with `.packages = c('nnet')`,
but its workers load the `FSSgam` namespace, and so used to load `gamm4` through
it. That stops as well, with no change of its own.

### Not established

What accounts for the residual 25 stalls in 140. #14 records one stall in 50 with
`mgcv` alone and a trivial body, so a smaller contributor was already known to
exist; whether the residual here is the same one was not shown. Whether any of
this is specific to WSL2 or to this `lme4`/`Matrix` build.

</details>


beckyfisher:
## Review round 1

An independent review of `59af807` found one blocking defect and two substantive
points. All three are addressed in `9c9938d`, two of them by withdrawing what the
pull request claimed rather than by defending it. The body has been rewritten.

## The suite required a Suggests package unconditionally

Eight test blocks build a `test.fit` with `MuMIn::uGamm(lme4 = TRUE)`, which
stops outright when `gamm4` is not installed. While `gamm4` was under `Imports`
that could not happen. `R CMD check` on the branch tarball with `gamm4` hidden
from the library path reported 2 ERRORs, the tests failing 8, and CRAN requires a
package to check successfully with its suggested packages absent. CI would not
have caught it: `setup-r-dependencies@v2` installs Suggests by default.

`skip_if_not_installed("gamm4")` is now in `fixture_coral_ugamm()` and at the one
direct `uGamm(lme4 = TRUE)` call. Verified against a library view built by
symlinking every package in the site library except `gamm4`: 547 passing, 0
failing, 0 errors, 14 skipped, against 571 and 7 with `gamm4` present.

## The gamm4 branch of worker_packages() is removed

`MuMIn::uGamm()` calls `require("gamm4")` itself when `lme4 = TRUE`, so a worker
given `.packages = c("mgcv", "MuMIn", "FSSgam")` refits a `uGamm(lme4 = TRUE)`
model set with no failures and ends with `gamm4` and `lme4` loaded. Naming
`gamm4` in `.packages` therefore bought no correctness, and on this change's own
account of the mechanism it only moved that load forward to the dispatch, which
is the condition the stall is associated with, for exactly the users who fit
`gamm4` models.

`worker_packages()` now takes no argument and returns
`c("mgcv", "MuMIn", "FSSgam")`, with the reason for each on the face of it.

## The trivial-body measurement is withdrawn

The review ran #14's own reprex and got 5 stalls in 40 with `gamm4` against 11 in
40 without, in the opposite direction, at a much lower overall rate than either
this pull request or #14 recorded.

The figure should not have been offered. Its two arms were run one after the
other rather than interleaved, so they are confounded with time and with machine
load, and a failure to reproduce them is the expected result rather than a
surprise. It is removed from `NEWS.md` and from the body.

What remains is the end-to-end comparison, 44 stalls in 140 against 25 in 140,
which was interleaved in blocks of ten and is the measurement the change now
rests on. It is one session's measurement of a rate on one host and `NEWS.md`
now says so.

An attempt to re-run the trivial-body comparison interleaved was abandoned rather
than reported. An unrelated job had taken this host's load average above 20 on 22
cores, and every run of both arms timed out, 10 of 10 each. A measurement in
which both arms saturate says nothing.

## Nits

The invariant test now checks `Depends`, `Imports` and `LinkingTo`, not `Imports`
alone. `worker_packages()` is computed inside the `parallel == TRUE` branch. The
prompt log has its `## Session:` header.

The `DESCRIPTION` bump to 1.1.0.9000 and the `cran-comments.md` note are kept.
They are byte-identical to the same change on #18 and #21, where the first review
raised the opposite point, that a development `NEWS.md` heading and a release
version should not disagree. Splitting them out would leave whichever of the
three merges first carrying the mismatch.

## Raised separately

#30, a `dsm` test.fit cannot be refitted under `parallel = TRUE`, because `dsm`
is never loaded onto the workers. Pre-existing; `worker_packages()` is the
natural place to fix it, which is why it is worth recording now rather than
later.

Suite: 571 passing, 0 failing, 7 skipped, against 568 on `master`.

beckyfisher:
## Review round 2

An independent review of `9c9938d` found no blocking defect, and reproduced every
non-timing claim exactly, including `R CMD check` on the tarball with `gamm4`
hidden returning Status OK. Three substantive points remained, all about what the
change is recorded as doing rather than what it does. All are corrected in
`f316a25`.

## The stated reason for keeping MuMIn was wrong

`MuMIn` registers an S3 method `update.gamm`, whose body calls `update.default`
from inside the `MuMIn` namespace, so `uGamm` is resolved there and never from
the worker's search path. A `uGamm(lme4 = TRUE)` model set refits 8 of 8
candidates with `.packages = "FSSgam"` alone.

`mgcv` genuinely is required: a `gam()` test.fit refits 0 of 4 without it,
failing with `could not find function "gam"`.

`MuMIn` stays in the vector, defensively, for a `test.fit` whose class carries no
such method. `R/utils.R` now says that rather than asserting a mechanism that does
not hold.

## The claim did not hold for gamm4 users

Unserialising a `uGamm(lme4 = TRUE)` test.fit loads `gamm4` and `lme4` on the
worker by itself, and `doSNOW` sends the export environment in the same
`clusterCall` that carries the package vector, so that happens before the
`library()` loop runs. The comment claiming that naming `gamm4` in `.packages`
"would only move that load forward to the dispatch" had it the wrong way round:
for those users the load is at the dispatch either way, and what the vector
controls is whether every other model set pays for it as well.

The 44-in-140 against 25-in-140 measurement was made on a `gam()` test.fit and
should not be read as extending to `gamm4` users. `NEWS.md` now states that as
one of three qualifications on the figure.

## The dependency change is user-visible and was not documented as one

A fresh `devtools::install_github()` no longer installs `gamm4` or `lme4`, so
anyone fitting through `MuMIn::uGamm(lme4 = TRUE)`, as the companion repository's
case study 2 does, now sees `cannot load package 'gamm4'` on a clean machine.
`NEWS.md` asserted instead that such a user "already has `gamm4` installed",
which is an assumption about the user rather than a statement of what changed.
The entry now leads with the actionable sentence.

## Also corrected

It was the `@importFrom gamm4 gamm4` directive, not the `Imports` field, that
made the `FSSgam` namespace load `gamm4`. The earlier wording implied that
`MuMIn`, `nnet` and `doSNOW` are also loaded on a worker that way, and they are
not. Both had to change together regardless, since an `Imports` entry with no
`NAMESPACE` import raises "All declared Imports should be used" — which is what
covers the half of the invariant the new test does not.

One unclaimed beneficiary is now recorded: `check_correlations(parallel = TRUE)`
dispatches with `.packages = c('nnet')`, but its workers load the `FSSgam`
namespace and so used to load `gamm4` through it. That stops as well.

## Raised separately

#31, `Depends: R (>= 3.5)` is not reachable, since `MuMIn` declares
`R (>= 4.4.0)`. That weakens the reason given for rejecting
`setup_strategy = "sequential"` as a mitigation, which is recorded on the issue.
#20 already covers the two redirecting URLs.

## On the evidence

The review recorded, unprompted, that this change is justified on dependency
hygiene alone and should stand even if the stall result never replicates. That is
the right reading. Nothing in `R/` calls `gamm4`, and requiring every user to
install `gamm4`, `lme4`, `Matrix`, `nloptr`, `Rcpp` and `reformulas` for a class
name test is not defensible on its own terms.

Suite: 571 passing, 0 failing, 7 skipped, against 568 on `master`; 547 passing,
0 failing, 0 errors, 14 skipped with `gamm4` hidden from the library path.

beckyfisher:
## Review round 3, and the withdrawn measurement remade

An independent review of `f316a25` found no blocking and no substantive defect.
The round 2 corrections were reproduced from private installs of both refs, and
the worker behaviour was established by simulating a worker in a fresh R process
rather than by running a cluster, so none of it depends on the host's state:
a `gam()` model set refits 8 of 8 with `mgcv` attached and 0 of 8 without, a
`uGamm(lme4 = TRUE)` set refits with nothing attached at all, and a doctored
build with `gamm4` back under `Imports` but no `NAMESPACE` import loads nothing,
confirming it was the directive.

One nit, now fixed in `1ea534f`: the test comment named the wrong check.
Re-adding the `@importFrom` while `gamm4` stays under `Suggests` raises
"Namespace dependency missing from DESCRIPTION Imports/Depends entries", an
ERROR; the NOTE the comment named catches the reverse. The comment now names both
directions and records that the expectation covers neither, only `gamm4` being
made a hard dependency again.

## The narrower measurement, remade

The unrelated job that had this host at a load average above 20 finished, and the
comparison withdrawn in round 1 was run again under the design the end-to-end one
uses, alternating run by run rather than taking each arm to completion in turn:

| workers attach | stalled |
|---|---|
| `mgcv`, `gamm4`, `MuMIn`, `FSSgam` | 34 of 50 |
| `mgcv`, `MuMIn`, `FSSgam` | 19 of 50 |

Chi-squared p = 0.005, `foreach()` body `i * 2` so that no package code runs, one
cluster per fresh R process. It agrees with the end-to-end figure in direction
and rough magnitude, and it is recorded in `NEWS.md` alongside it in `5d7f96e`.

The lesson is the design rather than the number. The original 23 in 60 against 0
in 60 had one arm run to completion before the other began, so it is confounded
with anything that changed on the host in between, and round 1's failure to
reproduce it is what that design predicts. Only alternating designs are quoted
now, on this pull request and in `NEWS.md`.

Neither figure extends to a `uGamm(lme4 = TRUE)` test.fit, for the reason given
in round 2.

Rounds 2 and 3 found no blocking defect and round 3 nothing substantive, so this
pull request is finished from my side.

Suite: 571 passing, 0 failing, 7 skipped, against 568 on `master`; 547 passing,
0 failing, 0 errors, 14 skipped with `gamm4` hidden from the library path, where
`R CMD check` returns Status OK.

beckyfisher:

Three rounds, each by a session with no prior context. Round 1 found one blocking
defect; rounds 2 and 3 found none, and round 3 nothing substantive. This is the
record of what was found and fixed.

## The change

`gamm4` moves from `Imports` to `Suggests` and is no longer named in the packages
the `doSNOW` workers attach. Nothing in the package calls it: the only uses are
`inherits(x, "gamm4")` class tests and an `@importFrom` directive for a function
never called, and that directive is what made every worker load `gamm4` and
`lme4`, for every model set whatever the family.

This is user-visible. A fresh install no longer brings `gamm4` or `lme4`, so
anyone fitting through `MuMIn::uGamm(lme4 = TRUE)` has to install `gamm4`
themselves. `NEWS.md` leads with that.

## Defects found during review and corrected here

**The test suite required a Suggests package unconditionally.** Seven blocks
build a `uGamm(lme4 = TRUE)` test.fit, which stops outright without `gamm4`.
`R CMD check` on the tarball with `gamm4` hidden reported 2 ERRORs. CRAN requires
a package to check successfully with its suggested packages absent, and CI would
not have caught it, since `setup-r-dependencies` installs Suggests by default.
Now skipped; the same check returns Status OK. Round 1.

**Three claims in the recorded account were wrong and are corrected.** An earlier
version named `gamm4` in the worker packages for a `gamm4` test.fit, on the
reasoning that such a refit needs it; it does not, and doing so restored the
condition the change identifies as harmful for exactly those users. The stated
reason for keeping `MuMIn` was wrong: a `uGamm` fit dispatches to MuMIn's own
`update.gamm`, whose body runs inside that namespace, so a `uGamm` model set
refits with `MuMIn` absent from the vector; `mgcv` genuinely is required, 0 of 8
without it. And it was the `@importFrom` directive rather than the `Imports`
field that made the namespace load `gamm4`. Rounds 1 to 3.

## On the evidence

Two measurements, both alternating the two arms run by run so that each sees the
same machine state, one cluster per fresh R process:

| measurement | before | after |
|---|---|---|
| `fit_model_set(parallel = TRUE)`, `case_study1`, 8 candidates | 44 stalls of 140 | 25 of 140 |
| `foreach()` body `i * 2`, no package code | 34 stalls of 50 | 19 of 50 |

Chi-squared p = 0.013 and p = 0.005. A reduction, not a removal; #14 stays open.
Neither extends to a `uGamm(lme4 = TRUE)` test.fit, since unserialising such an
object loads `gamm4` on the worker before the packages are attached.

An earlier figure taken with one arm run to completion before the other was
withdrawn after round 1 could not reproduce it, and was remade under the
alternating design. Only alternating designs are quoted.

The change stands on dependency hygiene without the stall result at all: nothing
in `R/` calls `gamm4`, and requiring every user to install `gamm4`, `lme4`,
`Matrix`, `nloptr`, `Rcpp` and `reformulas` for a class name test is not
defensible on its own terms.

## Raised separately

#30, a `dsm` test.fit cannot be refitted under `parallel = TRUE`. #31,
`Depends: R (>= 3.5)` is not reachable, since `MuMIn` requires `R (>= 4.4.0)`.

## State

571 passing, 0 failing, 7 skipped, against 568 on `master`; 547 passing, 0
failing, 0 errors, 14 skipped with `gamm4` hidden, where `R CMD check` returns
Status OK.

PR #32, opened 2026-09-03
Batch 1: redirecting URLs, the unguarded null fit, and the undocumented r2.vals.unique column

### LINKED ISSUE #19: check_non_linear_correlations() does not guard the intercept-only fit against try-error
## What

In `estimate_non_linear_correlation()` the fitted and intercept-only multinomial
models are both wrapped in `try()`, but only `fit` is tested before the two
deviances are used.

## Why it matters

If the intercept-only fit fails while the fitted model succeeds, `round()`
receives the character vector `try()` returns and raises "non-numeric argument
to mathematical function", which propagates out of
`check_non_linear_correlations()` rather than leaving the cell `NA`, which is
how a failed `fit` is already handled. `check_correlations()` was given this
guard in FSSgam_package#18, so the two sibling functions now differ.

## Evidence

`R/check-non-linear-correlations.R`, in `estimate_non_linear_correlation()`:

```r
null.fit=try(nnet::multinom(response.var1 ~ 1,trace=FALSE,data=dat.r)$deviance,silent=TRUE)
if(!inherits(fit,"try-error")){
   if(round(fit,4)==round(null.fit,4)){r.est=0}else{
  r.est=sqrt(1-(fit/null.fit))}
```

No reproduction is offered. Both models are fitted on the same `dat.r`, and
`multinom()`'s relevant failure ("need two or more classes to fit a multinom
model") depends only on the response, which the two models share, so a failed
null with a successful fit could not be constructed. The change is for
consistency with `check_correlations()` and to make the failure mode
inspectable, not to correct an observed error.

<details>

The change is the same one made in FSSgam_package#18:

```r
if(!inherits(fit,"try-error")&&!inherits(null.fit,"try-error")){
```

with a comment recording that the null half is defensive and that no case
reaching it is known, so a later session does not spend time looking for one.

There is no test to write that fails first. If one is wanted, it would have to
mock `nnet::multinom` to fail on the second call only, which pins the
implementation rather than the behaviour.

</details>

### LINKED ISSUE #20: DESCRIPTION URLs redirect, producing a CRAN incoming feasibility NOTE

### LINKED ISSUE #24: r2.vals.unique is undocumented, and its meaning depends on null.terms
## What

`fit_model_set()` returns `r2.vals.unique` in `mod.data.out`, and the column is
documented nowhere. The roxygen for `mod.data.out` lists AICc, BIC, delta values,
weights, "an estimate of the model R2", and the presence/absence indicator
columns. `r2.vals.unique` is not among them, and it appears in no `.Rd` file.

## Why it matters

The column is more useful than its absence from the documentation suggests, and
the name invites a reading it does not support.

It is computed as the model's R² minus the **null model's** R², so it is the
variance explained *relative to whatever was placed in `null.terms`*. With the
default `null.terms = ""` the null model is intercept-only and the column is the
model's R². With a forced term it becomes the variance explained over and above
that term, which is a different quantity — and the same column name means both.

The name reads as a per-*term* unique contribution, in the sense of a variance
partition. It is per-*model*. The two coincide only when `max.predictors = 1`,
where the column does give a clean per-candidate contribution and is arguably the
most directly interpretable number in the output. Above that it is the joint
contribution of whatever combination the model holds.

A user comparing `r2.vals.unique` across two runs with different `null.terms` is
comparing quantities with different baselines, with nothing in the output to
indicate it.

## Evidence

`R/fit-model-set.R`:

```r
null.r2 <- mod.data.out$r2.vals[which(mod.data.out$modname == "null")]
mod.data.out$r2.vals.unique <- mod.data.out$r2.vals - null.r2
```

and the roxygen describing the returned object, which does not mention it:

```
#' mod.data.out - A data.frame that contains the statistics associated with each
#' model fit. This includes AICc and BIC, delta values (e.g. AICc-(min(AICc)),
#' corresponding weight values (Burnham and Anderson 2003), an estimate of the
#' model R2, and a column for each of the included predictor variables ...
```

`grep -rn "r2.vals.unique" R/ man/` returns only the three lines that construct it.

<details>

### Suggested wording

Add to the `mod.data.out` description, alongside the existing R² sentence:

> `r2.vals.unique` — the model R² minus the R² of the null model, that is, the
> variance explained beyond the terms supplied in `null.terms`. Where
> `null.terms` is empty this equals `r2.vals`. Note that it is a per-model
> quantity: with `max.predictors` greater than one it is the joint contribution of
> all predictors in that model, not a variance partition among them. Values are
> comparable only within a model set sharing the same `null.terms`.

The `r2.type` argument should also be cross-referenced, since `r2.vals` and
therefore `r2.vals.unique` change meaning with it — deviance explained,
`summary()$r.sq`, or the R² of an observed-on-fitted linear model.

### Possibly also worth noting

Because the subtraction is on whatever scale `r2.type` produced, the difference is
not guaranteed to be non-negative for every fit: a candidate that fits worse than
the null on the chosen measure gives a negative value. Whether that is worth
mentioning depends on how often it is seen in practice.

### Definition of done

`?fit_model_set` describes the column, states its dependence on `null.terms`, and
says it is per-model rather than per-term.

</details>

beckyfisher:
## What

Three issues with no design decision in any of them: two redirecting `DESCRIPTION` URLs, an unguarded `try()` result, and an undocumented column of `mod.data.out`.

Closes #20, closes #19, closes #24.

## Why it matters

- **#20** is the only open issue producing an actual `R CMD check --as-cran` NOTE. CRAN submission is the stated end goal, and the incoming feasibility check runs on submission.
- **#19** leaves a failure mode that aborts `check_non_linear_correlations()` instead of leaving the cell `NA`, which is how the sibling failure is already handled. `check_correlations()` was given an equivalent guard in #16, so the two functions differed.
- **#24** leaves `r2.vals.unique` returned and described nowhere, under a name that reads as a per-term variance partition when it is a per-model quantity whose baseline is `null.terms`.

## Evidence

**#20.** `curl` against each URL, this host, 2026-09-03:

| URL | status |
|---|---|
| `https://beckyfisher.github.io/FSSgam_package` | 301 |
| `https://beckyfisher.github.io/FSSgam_package/` | 200 |
| `https://beckyfisher.github.io/FSSgam` | 301 |
| `https://beckyfisher.github.io/FSSgam/` | 200 |
| `https://github.com/beckyfisher/FSSgam_package` | 200, unchanged |

`R CMD check --no-manual --as-cran` on a tarball built from this branch, Debian WSL2, R 4.6.1, with network access, reports:

```
* checking CRAN incoming feasibility ... NOTE
Maintainer: 'Rebecca Fisher <r.fisher@aims.gov.au>'
New submission
Version contains large components (1.1.0.9000)
```

The URL block is gone. What remains is the development version string and the new-submission line, neither of which is #20.

**#19.** No reproduction. Both models are fitted on the same `dat.r`, and `multinom()`'s relevant failure depends only on the response, which the two share, so a failed null with a successful fit could not be constructed. The change is for consistency with `check_correlations()` and to make the failure mode inspectable.

**#24.** `grep -rn "r2.vals.unique" R/ man/` returned only the three lines constructing it.

## Verification

`devtools::check()` on this branch: 0 errors, 0 warnings, 0 notes. `testthat::test_local()`: 616 passing, 0 failing, 7 skipped blocks (Debian WSL2, R 4.6.1, testthat 3.3.2, 2026-09-03; the skips are the opt-in parallel tests). No test was added or changed, and none was expected to change: two of the three issues are documentation and the third has no constructible failing case.

<details>

### #20

The two `github.io` entries in `DESCRIPTION`'s `URL` field are rewritten with a trailing slash. `README.md` used the same `https://beckyfisher.github.io/FSSgam` URL and is corrected too. `man/FSSgam-package.Rd` follows from `devtools::document()`.

The `README.md` half is correctness rather than a check finding. It does ship in the tarball, but R's URL database does not extract that URL: `tools:::url_db_from_package_sources(".")` scans markdown link destinations, not plain-text URLs, so only the badges and the licence link in that file are checked.

Two URLs in the shipped tarball are still not 200, both predating this branch and neither producing a NOTE in the `--as-cran` run above: `https://CRAN.R-project.org/package=FSSgam` in `README.md` (the canonical badge form, expected to resolve on acceptance) and `https://onlinelibrary.wiley.com/doi/abs/10.1002/ece3.4134` in four `.Rd` files (403 to an automated request). Recorded so that a later `check_url_db()` run showing two hits is not read as a regression.

`README.md` uses CRLF line endings, so it was edited with a pattern anchored on `\r` to keep the diff to one line.

### #19

`R/check-non-linear-correlations.R`, in `estimate_non_linear_correlation()`:

```r
if(!inherits(fit,"try-error")&&!inherits(null.fit,"try-error")){
```

with a comment recording that the null half is defensive, why no case reaching it is known, and what the failure is without it, so a later session does not spend time looking for one.

No test. Constructing the case would require mocking `nnet::multinom` to fail on the second call only, which pins the implementation rather than the behaviour. This follows the issue's own reasoning rather than departing from it.

It is equivalent rather than identical: `check_correlations()` also tests `!is.null(null.fit)`, its `null.fit` being `NULL` by design in one branch.

**What this change does not do.** The `lm()` call in the same function, and the equivalent one in `check_correlations()`, are not wrapped in `try()` at all, and that path *is* reachable: a single-level factor predictor aborts both functions, and `generate_model_set()` with it. Raised as #33 and marked at both call sites. The `NEWS.md` entry states this, so the change is not read as making the function degrade to `NA` in every failure.

### #24

`full_subsets_gam()`'s `Value` section describes the column as well, not only `fit_model_set()`'s. #24's definition of done names only `?fit_model_set`, but leaving one of the two descriptions of `mod.data.out` silent about the column would have replaced one gap with a disagreement. Only `mod.data.out` passes through unaltered: `full_subsets_gam()` returns six elements against `fit_model_set()`'s four.

The issue's evidence quotes the subtraction without its surrounding `if(report.unique.r2==TRUE)`, so the wording was written against the code rather than the issue text: the column is present in `mod.data.out` either way, and is `NA` unless `report.unique.r2` is `TRUE`. `extract_mod_dat()` initialises it in the list it returns.

Added to `fit_model_set()`'s `@return` description of `mod.data.out`, and the `report.unique.r2` `@param` in both `fit_model_set()` and `full_subsets_gam()` rewritten to name the column it populates.

The four statements the documentation now makes are: the baseline is `null.terms`; the quantity is per-model, not per-term, and per-predictor only at `max.predictors = 1`; it is on the `r2.type` scale and can be negative; and it is comparable only within a set sharing both.

It also states what an empty `null.terms` means, since that is when the column equals `r2.vals`. `build_null_model()` updates the `test.fit` with the formula `~1`, which drops every term of it, a random effect written `s(site, bs = 're')` included — that is what `null.terms` exists to put back. A random structure supplied as a separate argument, as in `uGamm(random = ~(1|site))`, is not part of the formula and is retained either way. Both were measured on this branch.

`R/full-subsets-gam.R` uses CRLF line endings and was edited so as to keep them.

### Not done here

Nothing was deferred from these three issues. One defect found while reviewing this branch is deferred to #33, for the reason given above.

</details>


beckyfisher:
Three independent review rounds against this branch. No round found a defect in the package's behaviour. The findings were documentation inaccuracies, and the pull request body has been updated as each was corrected.

## What the reviews found

**Round 1.** No substantive defect. Four corrections: `full_subsets_gam()`'s own `Value` section still omitted `r2.vals.unique`, so the two documented descriptions of the same data.frame disagreed; the claim "the null model is intercept-only" is not right for a `uGamm` fit with a `random =` argument; "the same guard" as `check_correlations()` is not literal, since that function also tests `!is.null(null.fit)`; and a bare `#16` in a commit message needed repository qualification, `beckyfisher/FSSgam#16` also existing.

**Round 2.** Two substantive findings, both in text added in response to round 1.

The first correction had over-generalised in the opposite direction. `build_null_model()` updates the `test.fit` with the formula `~1`, and `update.formula()` drops every term of it — so a random effect written `s(site, bs = 're')` is *not* retained by the null model. That is the case `fit_model_set()`'s own `@examples` builds, and it is why `null.terms` exists. What is retained is a random structure supplied through a separate argument, `uGamm(random = ~(1|site))`, which is not part of the formula. Measured both ways; the two baselines differ by 0.13574 on `case_study1`.

"`full_subsets_gam()` returns the object `fit_model_set()` built, unaltered" was false. It returns six elements against `fit_model_set()`'s four, and its own `Value` block lists two of the extra ones, so the page contradicted itself. Only `mod.data.out` passes through.

**Round 3.** No substantive defect. Both round 2 statements were re-derived from scratch across four `test.fit` classes and hold. Five minor points were acted on: the `r2.vals.unique` text stated a necessary condition ("NA unless `report.unique.r2` is TRUE") that a reader can take for a sufficient one, the column being `NA` for a `gamm` fit under the default `r2.type` as well (#6); three constructions parent `CLAUDE.md` §12 rules out, all of which render onto the help page; a comment locating the `multinom()` fits "below", true of the file but not of the function; four lines of issue-number provenance that belong to the issue thread rather than to `R/`; and this body correcting itself across the fold, which §14 forbids.

## Raised as separate issues

Two pre-existing defects were found by these reviews, neither in scope here, both confirmed before filing.

**#33** — a single-level factor predictor aborts `check_correlations()`, `check_non_linear_correlations()` and `generate_model_set()`. The `lm()` call in each correlation function is not wrapped in `try()` at all, unlike every other fit in them, so the error propagates instead of leaving the cell `NA`. `data.frame(x = rnorm(20), f = factor(rep("a", 20)))` is enough. Both call sites are marked, and the `NEWS.md` entry for #19 states what the guard does not do, so it is not read as making the function degrade gracefully in every failure.

**#34** — an `mgcv::gamm()` fit cannot be a `test.fit` at all. It records no `call`, so `stats::update()` fails before any candidate is enumerated. This matters because `generate_model_set()`'s `null.terms` documentation directs the user to `gamm` for a correlation structure, its `@details` contrast `gam` against `gamm` at length, `r2.type` names `gamm` among the classes it reports for, and `extract_mod_dat()` has a `gamm` branch. The error message then advises the user to stop using `uGamm`, which is what they would have needed to use.

## State

`devtools::check()`: 0 errors, 0 warnings, 0 notes. `testthat::test_local()`: 616 passing, 0 failing, 7 skipped blocks — unchanged from `dev`, and no test was expected to change, two of the three issues being documentation and the third having no constructible failing case. `R CMD check --no-manual --as-cran` on a built tarball with the remote incoming checks enabled: one NOTE, stating the new submission and the development version string, with no invalid-URL block. Measured on Debian WSL2, R 4.6.1, testthat 3.3.2, 2026-09-03.

PR #35, opened 2026-09-03
Batch 2: raise the declared R floor to 4.4.0, and check it

### LINKED ISSUE #31: Depends: R (>= 3.5) is not reachable, since MuMIn requires R (>= 4.4.0)
## What

`DESCRIPTION` declares `Depends: R (>= 3.5)`, but `MuMIn`, a hard `Imports`
dependency, currently declares `Depends: R (>= 4.4.0)`. The declared floor is
therefore not reachable with current dependency versions.

## Why it matters

The floor shapes the code and is used to reject alternatives. The test suite
avoids `deparse1()` because it arrived in R 4.0.0, and
`parallel::makePSOCKcluster(setup_strategy = "sequential")` was rejected as a
mitigation for FSSgam_package#14 partly for the same reason. Neither constraint
buys anything if no one can install the package on R 3.5.

Raising the floor is not obviously right either: it would be a breaking change
for anyone pinned to an older `MuMIn`, and CRAN prefers the lowest floor a
package genuinely supports. What is wrong is that the two are not reconciled and
nothing checks the claim.

## Evidence

`utils::packageDescription("MuMIn")$Depends` is `R (>= 4.4.0)` at MuMIn 1.48.19.
No job in `.github/workflows/R-CMD-check.yaml` builds against R 3.5; the oldest
is `ubuntu-latest` with `r: 'release'`.

<details>

Two decisions, in order.

First, whether the floor should track the hard dependencies. If it should, it
becomes `R (>= 4.4.0)` while `MuMIn` requires that, and the notes in
`tests/testthat/helper-fixtures.R` about `deparse1()` and `deparse_one()` can go,
as can the `setup_strategy` reasoning on FSSgam_package#14.

Second, whether anything should check it. `R CMD check` does not verify that a
declared floor is consistent with a dependency's, so a comment in `DESCRIPTION`
recording where the number comes from is the practical option.

Found while reviewing FSSgam_package#29.

</details>

beckyfisher:
Stacked on #32. Review that first; this pull request's diff against `dev` will include it until #32 merges.

## What

`DESCRIPTION` declares `Depends: R (>= 4.4.0)`, raised from `R (>= 3.5)`.

Closes #31.

## Why it matters

The old floor was not reachable, so it was a claim about support that no one could act on and nothing checked. Two hard `Imports` require R 4.4.0 in their own right, so the package could not be installed on anything lower whatever `DESCRIPTION` said.

The floor is not inert. It shapes the code and is used to reject alternatives: the test suite avoided `deparse1()` on account of it, and `parallel::makePSOCKcluster(setup_strategy = "sequential")` was rejected as a mitigation for FSSgam_package#14 partly for the same reason. Neither constraint provided any benefit.

Raising it is nonetheless a change a user can feel. Someone pinned to an older `MuMIn` and `mgcv` may have a working configuration that the new floor excludes, which is why `NEWS.md` states it rather than treating it as a correction.

## Evidence

Declared `Depends` of every direct dependency, measured on this host, R 4.6.1, 2026-09-03. Transitive dependencies were checked separately and none exceeds 4.4.0:

| package | role | `Depends` |
|---|---|---|
| `MuMIn` 1.48.19 | Imports | `R (>= 4.4.0)` |
| `mgcv` 1.9-4 | Imports | `R (>= 4.4.0)`, `nlme (>= 3.1-64)` |
| `doSNOW` 1.0.20 | Imports | `R (>= 2.5.0)`, … |
| `foreach` 1.5.2 | Imports | `R (>= 2.5.0)` |
| `nnet` 7.3-20 | Imports | `R (>= 3.0.0)`, … |
| `gamm4` 0.2-6 | Suggests | `R (>= 2.9.0)`, … |
| `testthat` 3.3.2 | Suggests | `R (>= 4.1.0)` |
| `covr` 3.6.5 | Suggests | `R (>= 3.1.0)`, … |

The binding constraint is 4.4.0, from `MuMIn` **and** `mgcv`. The issue names only `MuMIn`; `mgcv` was found while checking it, and matters because it means the floor does not move if `MuMIn` alone changes.

## The second decision the issue poses

The issue asks whether anything should check the floor, noting that `R CMD check` does not verify a declared floor against a dependency's, and proposing a comment in `DESCRIPTION` as the practical option. Neither the comment nor a CI matrix entry answers it, so this pull request adds a check that does.

**`tools/check-r-floor.R`, run in CI as a new `r-floor-consistency` job.** It reads `DESCRIPTION`'s declared floor, walks the transitive closure of `Depends`, `Imports` and `LinkingTo`, reads each package's own declared R floor as resolved for that job, and fails if the declared floor is lower than the highest of them. Output on this branch:

```
DESCRIPTION declares R (>= 4.4.0)

package        version    direct   declared R floor
codetools      0.2-20     -        2.1
doSNOW         1.0.20     yes      2.5.0
foreach        1.5.2      yes      2.5.0
insight        1.5.2      -        3.6
iterators      1.0.14     -        2.5.0
lattice        0.22-7     -        4.0.0
Matrix         1.7-4      -        4.4
mgcv           1.9-4      yes      4.4.0
MuMIn          1.48.19    yes      4.4.0
nlme           3.1-168    -        3.6.0
nnet           7.3-20     yes      3.0.0
snow           0.4-4      -        2.13.1

12 hard dependencies, 5 of them direct. Highest floor: R (>= 4.4.0), from Matrix, mgcv, MuMIn

OK: the declared floor is reachable.
```

**The closure, not the direct `Imports`.** `Matrix` declares `R (>= 4.4)` and is not a direct dependency of this package, arriving through `mgcv` and `MuMIn`, so a check of the direct `Imports` alone misses one of the three packages at the binding floor. For a package whose only route to `Matrix` is transitive it would report a floor reachable that is not.

**The script stops rather than reporting success wherever it cannot count a constraint.** Three such cases: a dependency that is not installed, an R constraint written in a form it does not parse, and a `DESCRIPTION` with no floor. Each was found passing silently during review — the same failure the script exists to catch, recurring inside the script itself — so each now fails loudly and names what it could not read, and every guard runs before any exit path.

and on a `DESCRIPTION` reverted to the state #31 reported, it exits 1:

```
Error: DESCRIPTION declares R (>= 3.5) but the hard dependencies require
R (>= 4.4.0). The declared floor is not reachable: raise it, or drop the
dependency that imposes the higher one.
```

**An `ubuntu-latest`, `R 4.4.0` matrix entry.** The package is built, installed and tested on an R at the declared floor for the first time.

**These check different things, and only the first sees what was wrong here.** Installing a package rejects a floor set *above* the running R and accepts any floor *below* it. So no matrix of R versions can detect a floor that is too low, which is the whole of #31. Measured directly on this host: with `DESCRIPTION` reverted to `R (>= 3.5)`, `R CMD INSTALL` and `devtools::check()` both succeed on R 4.6.1. No CI run was made against a reverted `DESCRIPTION`; the inference to the matrix jobs is from that local result and from what installation does, not from a green CI run.

The matrix entry names 4.4.0 literally rather than `oldrel-2`. `oldrel-2` resolves to 4.4.3 today and drifts as R advances, so a base function added in 4.4.1 would pass CI and fail at the declared floor.

**`DESCRIPTION` records no provenance comment.** Writing R Extensions §1.1.1 states the format does not support `#` comments — `read.dcf()` skips them as an implementation detail, not by specification — and any `desc`-based write deletes them silently, including a version bump and `roxygen2` updating `Config/roxygen2/version`. This repository has run both. The script holds the provenance instead, where it is executed rather than trusted.

<details>

### Consequences carried out with the change

`deparse_one()` in `tests/testthat/helper-fixtures.R` existed only because `deparse1()` arrived in R 4.0.0. It is deleted, and its 31 call sites across four test files now call `deparse1()` directly. The two are identical in behaviour: R's `deparse1()` is `paste(deparse(expr, width.cutoff), collapse = collapse)` with `width.cutoff = 500L` and `collapse = " "`, which is what `deparse_one()` was. The suite is unchanged at 616 passing.

`CLAUDE.md`'s note recording the old constraint is rewritten to state the surviving reason for not using bare `deparse()` — its default `width.cutoff` of 60 wraps 8 of 39 formulas, and a wrapped formula deparses to a different string — without the R 3.5 justification. Its `Depends` line is updated with the provenance.

Its `Imports`/`Suggests` list is corrected while adjacent: it still named `gamm4` under `Imports`, which stopped being true when `gamm4` moved to `Suggests`.

### Not done here

`parallel::makePSOCKcluster(setup_strategy = "sequential")` is now available as a mitigation to test against FSSgam_package#14, having been rejected partly on the old floor. That belongs with #14 itself and is not attempted here; it is recorded as a comment on that issue.

`planning/golden-master/capture.R` still defines its own local `deparse_one()`. Left alone: it is a self-contained record of what was run, and its copy still works.

### The script's edge cases

It is the only new code here, so it was run against twelve constructed inputs, three of them requiring scratch packages installed for the purpose. The floor written `R (>=4.4.0)` and `R(>= 4.4.0)` without the usual spaces; `R (>= 4.4)` against `R (>= 4.4.0)`, which must compare equal and do, `package_version()` padding; `R (>= 4.10.0)` against `R (>= 4.9.0)`, which string comparison orders wrongly and `package_version()` does not; version constraints on the `Imports` entries; a hard dependency declared in `Depends` rather than `Imports`, which must not disable the check and does not; a `DESCRIPTION` with no R floor at all; an uninstalled dependency; and package names beginning with R — `Rcpp`, `R6` — which must not be read as the R entry and are not, the pattern being anchored on `R` followed by `(`.

Three scratch dependencies were installed to reach the constraint-parsing paths: one declaring `R (> 4.5.9)`, which is counted as a floor of 4.5.9 and correctly fails a declared 3.5; one declaring `R (<= 9.9.0)`, which stops the script naming the package and the constraint rather than discarding it; and one declaring `Depends: methods`, which reaches the "nothing to check" path legitimately.

### Verification

`devtools::check()`: 0 errors, 0 warnings, 0 notes. `testthat::test_local()`: 616 passing, 0 failing, 7 skipped blocks, unchanged. `tools/check-r-floor.R` was run against 23 constructed cases covering every exit path. The workflow file parses as YAML with the two jobs and the five matrix entries expected. Debian WSL2, R 4.6.1, 2026-09-03.

All six jobs pass at the current head — run 33726514953 on `381a3b3`: `declared R floor is reachable` and `ubuntu-latest (4.4.0)`, alongside the four pre-existing matrix jobs. That confirms two things this host cannot: that `r-lib/actions` resolves a literal `4.4.0`, and that the script runs correctly against a dependency set installed by `setup-r-dependencies` rather than this machine's.

</details>

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -12,3 +12,4 @@
 ^pkgdown$
 ^prompts$
 ^planning$
+^tools$
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -23,6 +23,19 @@ jobs:
           - {os: windows-latest, r: 'release'}
           - {os: ubuntu-latest,   r: 'devel', http-user-agent: 'release'}
           - {os: ubuntu-latest,   r: 'release'}
+          # Establishes that the package builds, installs and passes its
+          # tests on an R at the declared floor, which nothing did before.
+          # Named literally rather than as oldrel-2, which is 4.4.3 and
+          # drifts: a base function added in 4.4.1 would pass CI and fail at
+          # the floor. Update this whenever DESCRIPTION's Depends changes --
+          # the floor-consistency job below fails if the two disagree in the
+          # direction that matters.
+          #
+          # Note what this job cannot do. Installing rejects a floor set
+          # ABOVE the running R and accepts any floor BELOW it, so no matrix
+          # of R versions detects a floor that is too low, which is the whole
+          # of FSSgam_package#31. That is the floor-consistency job's work.
+          - {os: ubuntu-latest,   r: '4.4.0'}
 
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
@@ -47,3 +60,39 @@ jobs:
         with:
           upload-snapshots: true
           build_args: 'c("--no-manual","--compact-vignettes=gs+qpdf")'
+
+  # The check R CMD check cannot make: whether DESCRIPTION's declared
+  # "Depends: R (>= x)" is reachable given what the hard dependencies
+  # themselves require, transitively. Installing accepts any floor below the
+  # running R, so the matrix above is blind to a floor set too low -- which is
+  # exactly what FSSgam_package#31 reported, and what left "R (>= 3.5)"
+  # standing while mgcv and MuMIn both required R (>= 4.4.0).
+  #
+  # Run against the versions resolved for this job, so a dependency raising
+  # its own floor fails this job rather than going unnoticed until a user
+  # cannot install the package. The script stops if any hard dependency is
+  # not installed, rather than reporting a floor derived from part of the
+  # closure.
+  r-floor-consistency:
+    runs-on: ubuntu-latest
+    name: declared R floor is reachable
+
+    env:
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: r-lib/actions/setup-r@v2
+        with:
+          use-public-rspm: true
+
+      # The script reads each dependency's own DESCRIPTION, so they have to be
+      # installed, and it stops rather than passing if any is not. needs: check
+      # installs more than it reads -- Suggests included -- which is harmless.
+      - uses: r-lib/actions/setup-r-dependencies@v2
+        with:
+          needs: check
+
+      - name: Compare the declared floor against the dependency closure
+        run: Rscript tools/check-r-floor.R
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -59,10 +59,52 @@ the permanent, citable reference and must not be modified.
 Dependencies are defined in `DESCRIPTION` (Imports + Suggests). Read that file
 to determine what is available — don't assume this list stays in sync.
 
-- **Imports**: `doSNOW`, `foreach`, `gamm4`, `mgcv`, `MuMIn`, `nnet`,
-  `parallel`, `stats`, `utils`
-- **Suggests**: `covr`, `testthat (>= 3.1.5)`
-- **Depends**: `R (>= 3.5)` only — no other package may go in `Depends`
+- **Imports**: `doSNOW`, `foreach`, `mgcv`, `MuMIn`, `nnet`, `parallel`,
+  `stats`, `utils`
+- **Suggests**: `covr`, `gamm4`, `testthat (>= 3.1.5)`. `gamm4` moved out of
+  `Imports` so that a parallel worker no longer loads it and `lme4` with it
+  for every model set (FSSgam_package#14).
+- **Depends**: `R (>= 4.4.0)` only — no other package may go in `Depends`.
+  The floor is not chosen: `MuMIn` and `mgcv` both declare
+  `Depends: R (>= 4.4.0)`, so nothing lower is reachable.
+
+  Two jobs in `.github/workflows/R-CMD-check.yaml` cover it, and they check
+  different things. The `r: '4.4.0'` matrix entry establishes that the
+  package installs and passes on an R at the declared floor. The
+  `r-floor-consistency` job runs `tools/check-r-floor.R`, which compares the
+  declared floor against the `Depends` of every hard dependency, walking the
+  closure rather than the direct `Imports`, and fails if the declared one is
+  lower. The closure matters: `Matrix` declares `R (>= 4.4)` and arrives only
+  through `mgcv` and `MuMIn`, so a check of the direct `Imports` misses it.
+
+  The script stops rather than reporting success in three cases where it
+  cannot count a constraint: a dependency that is not installed, an R
+  constraint in a form it does not parse, and a `DESCRIPTION` with no floor.
+  Each was added after review found it in turn silently passing, and a fourth
+  round found the guards themselves ordered wrongly — an empty-closure exit
+  placed above the uninstalled-dependency check accepted the declared floor,
+  having read nothing, for a package all of whose dependencies were missing.
+  **Anything added here must fail loudly rather than contribute no floor, and
+  must be placed above the reporting, not below it** — that is the failure
+  mode the script exists to catch, and it recurred four times inside the
+  script itself.
+
+  **Only the second detects the defect FSSgam_package#31 reported.**
+  Installing rejects a floor set above the running R and accepts any floor
+  below it, so no matrix of R versions sees a floor that is too low.
+  Measured 2026-09-03 on this host: `R CMD INSTALL` and `devtools::check()`
+  both succeed with `Depends: R (>= 3.5)` on R 4.6.1. No CI run was made
+  against a reverted `DESCRIPTION`; that every matrix job would stay green
+  follows from what installation does, and is an inference rather than a
+  measurement. Do not add an R version to the matrix and take it for a floor
+  check.
+
+  `DESCRIPTION` records no provenance comment. Writing R Extensions §1.1.1
+  states the format does not support `#` comments, and any `desc`-based write
+  deletes them silently -- `usethis::use_package()`, a version bump, and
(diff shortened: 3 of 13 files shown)


beckyfisher:
Five independent review rounds. The floor value was correct from the first round and never changed. Every defect found was in the mechanisms added to record and check it, and four consecutive rounds found the same one.

## The defect, four times

**A check that reports success without having checked.** It appeared in four different places, each fix exposing the next.

**Round 1 — the CI matrix entry.** The original change added an `oldrel-2` job and four documents said it checked the declared floor. It could not. Installing a package rejects a floor set *above* the running R and accepts any floor *below* it, so an R 4.4 job detects a floor set too high and is blind to a floor set too low — which is the whole of #31. With `DESCRIPTION` reverted to `R (>= 3.5)`, `R CMD INSTALL` and `devtools::check()` both succeed on R 4.6.1.

Replaced with `tools/check-r-floor.R`, which compares the declared floor against what the dependencies themselves require. The matrix entry is kept, because building and testing at the floor is worth doing and nothing did it before, and is now described as that.

**Round 2 — the script's closure walk.** It read only the direct `Imports`. `Matrix` declares `R (>= 4.4)` and is not a direct dependency of this package, arriving through `mgcv` and `MuMIn`. And an uninstalled dependency contributed no floor and left the job green — the state `CLAUDE.md` §3 records for `doSNOW` in fresh environments here. The script now walks the transitive closure and stops on a dependency it cannot read.

**Round 3 — the constraint parser.** It returned `NULL` both for "declares no R constraint" and for "declares one I cannot read", accepting only `>=`. A dependency declaring `Depends: R (<= 9.9.0)` produced "No hard dependency declares an R floor; nothing to check" and exit 0. The parser now returns three states rather than two and stops on anything it cannot read.

**Round 4 — the order of the guards.** Each guard was correct; their order was not. The empty-closure exit added in round 3 sat above the uninstalled-dependency check added in round 2, so a package with every hard dependency missing reported "no hard dependencies outside base R" — while declaring two — and accepted the floor having read nothing. Every guard now runs before any exit path.

**Round 5 — no fifth instance.** Twenty constructed inputs across every exit path, all correct. One approximation was removed rather than documented: `R (> x)` was being read as `R (>= x)`, which a declared floor of exactly `x` satisfies when it should not. The operator is now recorded and each constraint tested as written.

## What the change is

`DESCRIPTION` declares `Depends: R (>= 4.4.0)`, raised from `R (>= 3.5)`. `mgcv` and `MuMIn`, both hard `Imports`, each declare `R (>= 4.4.0)`, so the old floor was not reachable and the declaration now states what was already true. `Matrix` declares `R (>= 4.4)` transitively.

Two CI jobs cover it and they check different things. `r-floor-consistency` runs the script; the `ubuntu-latest` `R 4.4.0` matrix entry builds and tests at the declared floor. Only the first detects the direction that went wrong here, and `CLAUDE.md` records that, so an R version is not added to the matrix and taken for a floor check.

`DESCRIPTION` records no provenance comment. Writing R Extensions §1.1.1 states the format does not support them, and any `desc`-based write deletes them silently — a version bump and `roxygen2` updating `Config/roxygen2/version` both do, and this repository has run both. The script holds the provenance instead, where it is executed rather than trusted.

`deparse_one()` in `tests/testthat/helper-fixtures.R` existed only because `deparse1()` postdated the old floor. Removed, its 31 call sites replaced.

## State

All six jobs pass at the head — run 33726514953 on `381a3b3`:

```
declared R floor is reachable: success
ubuntu-latest (4.4.0):         success
ubuntu-latest (devel):         success
ubuntu-latest (release):       success
windows-latest (release):      success
macos-latest (release):        success
```

`devtools::check()`: 0 errors, 0 warnings, 0 notes. `testthat::test_local()`: 616 passing, 0 failing, 7 skipped blocks, unchanged from `dev` throughout. Debian WSL2, R 4.6.1, 2026-09-03.

A comment on #14 records that raising the floor lifts the constraint on its `setup_strategy` mitigation. That direction remains uninvestigated and belongs to that issue.

Issue #33, opened 2026-09-03
A single-level factor predictor aborts both correlation functions and generate_model_set() through an unguarded lm()


beckyfisher:
## What

The `stats::lm()` call in `estimate_non_linear_correlation()`, and the equivalent
one in `check_correlations()`, are not wrapped in `try()`. Where the fit fails,
the error propagates out of the correlation function and aborts the whole call,
rather than leaving that cell `NA`, which is how a failed `multinom()` or `gam()`
in the same functions is already handled.

The commonest way to reach it is a factor predictor with one level.

## Why it matters

A single-level factor is an ordinary thing to end up with — a `ZONE` column that
survives a subset to one zone, a treatment column in a pilot subset. Nothing about
it is invalid input to `generate_model_set()` in principle: the correct outcome is
that the predictor explains nothing, and the cell is `NA` or zero.

Instead `generate_model_set()` stops with

```
Error in `contrasts<-`(`*tmp*`, value = contr.funs[1 + isOF[nn]]) :
  contrasts can be applied only to factors with 2 or more levels
```

which names neither the function, nor the argument, nor the predictor at fault.
A user with twenty predictors is told a factor somewhere has one level, and has
to find which.

## Evidence

Measured on `dev` at `656ef62` plus the branch of FSSgam_package#32, Debian WSL2,
R 4.6.1, mgcv 1.9-4, nnet 7.3-20, 2026-09-03.

A single-level factor, no missing values:

```r
d <- data.frame(x = rnorm(20), f = factor(rep("a", 20)))
check_correlations(d)             #> Error: contrasts can be applied only to ...
check_non_linear_correlations(d)  #> Error: contrasts can be applied only to ...
```

and through the public entry point:

```r
d2 <- FSSgam::case_study1
d2$onelev <- factor("a")
test.fit <- mgcv::gam(Herbivore.abundance ~ s(depth, k = 3, bs = "cr"),
                      family = mgcv::tw(), data = d2)
generate_model_set(use.dat = d2, test.fit = test.fit, pred.vars.cont = "depth",
                   pred.vars.fact = "onelev", max.predictors = 2, k = 3)
#> Error in `contrasts<-`(`*tmp*`, value = contr.funs[1 + isOF[nn]]) :
#>   contrasts can be applied only to factors with 2 or more levels
```

The asymmetry within one function is the clearest statement of the defect. On a
frame whose factor and continuous columns have no complete rows in common, the
[...]

PR #36, opened 2026-09-03
Batch 3: a 1x1 supplied cor.matrix, and an NA between two predictors

### LINKED ISSUE #26: A 1x1 supplied cor.matrix is silently discarded and recomputed
## What

`generate_model_set()` ignores a user-supplied `cor.matrix` that has exactly one
cell, and computes the matrix from `use.dat` instead. Nothing is reported.

## Why it matters

A model set with a single predictor is the one case where the supplied matrix is
1x1, and it is silently not used. The supplied value does not appear in
`predictor.correlations` and governs no screen, which is the opposite of the
documented contract that a supplied matrix replaces the automatic estimate.

## Evidence

On `master` at `ed50981`, using `case_study1` with `site` as a factor and a
Gaussian `gam()` test fit:

```r
cm <- matrix(0.5, 1, 1, dimnames = list("ZONE", "ZONE"))
model.set <- generate_model_set(use.dat = use.dat, test.fit = test.fit,
                                pred.vars.cont = NA, pred.vars.fact = "ZONE",
                                null.terms = "s(site,bs='re')",
                                max.predictors = 1, k = 3, cor.matrix = cm)
model.set$predictor.correlations
#      ZONE
# ZONE    1
```

The supplied 0.5 is gone.

<details>

`build_predictor_correlation_matrix()` in `R/generate-model-set.R` branches on
`if(length(cor.matrix)==1)`, using length as the sentinel for the `NA` default. A
1x1 matrix has length 1 and so takes the compute branch.

The test should be on the default value rather than on length, for instance
`length(cor.matrix)==1 && !is.matrix(cor.matrix) && !is.data.frame(cor.matrix)`.
`generate_model_set()`, `full_subsets_gam()` and the `@param` text all describe
the default as `NA`, so the sentinel can be written as such.

A regression test belongs in `test-generate_model_set.R` beside the existing
supplied-`cor.matrix` blocks: a single-predictor set with a 1x1 supplied matrix
whose value is not what `check_correlations()` would give, asserting that
`predictor.correlations` is the supplied matrix.

Found while reviewing FSSgam_package#21. Reproduces unchanged on that branch.

</details>

### LINKED ISSUE #27: An NA in a supplied cor.matrix fails with a message naming nothing
## What

Where a user-supplied `cor.matrix` carries `NA` in a cell between two
predictors, `generate_model_set()` stops with `missing value where TRUE/FALSE
needed` as soon as a candidate containing that pair is enumerated. The message
names neither the matrix, the argument, nor the pair.

## Why it matters

`NA` is a plausible entry in a hand-built matrix, for a pair no estimate was
available for. The failure arrives partway through model set construction and
gives the user nothing to act on. Whether it happens at all depends on an
unrelated argument: at `max.predictors = 1` the same matrix is accepted, because
the pair is never enumerated.

## Evidence

On `master` at `ed50981`, with `cm["depth","complexity"] <- NA` and
`max.predictors = 2`:

```
Error: missing value where TRUE/FALSE needed
```

<details>

Raised inside `combine_uncorrelated()` (`R/generate-model-set.R`) at
`if(max(abs(cor.mat.m[upper.tri(cor.mat.m)]))>cov.cutoff)`. The comment above
that function records the `if (NA)` behaviour as deliberate, and it is the right
behaviour there: the alternative silently drops the combination.

The check belongs one level up, in `build_predictor_correlation_matrix()`'s
supplied branch, beside the missing-predictor check, reporting any `NA` cell
between two predictors by naming its row and column. That leaves
`combine_uncorrelated()` unchanged and keeps the two branches consistent, since
the computed branch already replaces every `NA` with zero and so can never reach
it.

Note the interaction with FSSgam_package#15: `augment_supplied_correlation_matrix()`
deliberately leaves a user-supplied `NA` alone rather than inventing a value, so
under this change it would become a reported error rather than a silent one.

Found while reviewing FSSgam_package#21. Reproduces unchanged on that branch.

</details>

beckyfisher:
Stacked on #35, which is stacked on #32. Review those first; this pull request's diff against `dev` includes both until they merge.

## What

Two defects in how `generate_model_set()` handles a user-supplied `cor.matrix`, both found while reviewing #21.

Closes #26, closes #27.

## Why it matters

**#26.** A model set with a single predictor is the one case in which a supplied matrix is 1x1, and in exactly that case the matrix was discarded and recomputed from `use.dat`. The supplied value governed no screen and did not appear in `predictor.correlations` — the opposite of the documented contract that a supplied matrix replaces the automatic estimate. Nothing was reported. The sentinel had two copies, in `build_predictor_correlation_matrix()` and in the `factor_correlations()` helper inside `resolve_factor_interactions()`; both are replaced.

**#27.** An `NA` is a plausible entry in a hand-built matrix, for a pair no estimate was available for. It either stopped the call with `missing value where TRUE/FALSE needed`, which names neither the matrix, the argument, nor the pair, or was accepted silently — and which of the two happened depended on `max.predictors`, an unrelated argument.

## Evidence

**#26**, on the current branch point, `case_study1` with `site` as a factor and a Gaussian `gam()` test fit:

```r
cm <- matrix(0.5, 1, 1, dimnames = list("ZONE", "ZONE"))
model.set <- generate_model_set(use.dat = use.dat, test.fit = test.fit,
                                pred.vars.cont = NA, pred.vars.fact = "ZONE",
                                null.terms = "s(site,bs='re')",
                                max.predictors = 1, k = 3, cor.matrix = cm)
model.set$predictor.correlations
#      ZONE          before            #      ZONE          after
# ZONE    1                            # ZONE  0.5
```

**#27**, with `cm["depth","complexity"] <- NA`:

```
before, max.predictors = 2:  Error: missing value where TRUE/FALSE needed
before, max.predictors = 1:  accepted, model set built against a matrix with a hole in it

after:  Error: The correlation matrix has NA between terms screened against
               cov.cutoff: depth/complexity. If you supplied cor.matrix, supply
               a correlation for each of these pairs, or remove the predictor.
```

## Behaviour changes

Three, all recorded in `NEWS.md` rather than presented as corrections, because each is a call that used to run.

**A `cor.matrix` that is a length-one value of any class** — a string, or a list holding a matrix — previously satisfied the sentinel and was silently treated as nothing supplied, so the model set was built from a computed matrix as though the argument had not been given. It now errors naming the class.

**An `NA` on a pair that is actually screened** now errors with a message naming the pair, where it previously stopped with `missing value where TRUE/FALSE needed`.

**An `NA` on a pair that is not screened is accepted, and this departs from what #27 asked for.** The issue asked that the report not depend on `max.predictors`. It does, and on the interaction arguments too, because the report covers exactly the pairs some screen compares. Rejecting an inert cell would mean failing over something nothing reads.

Note that "compared" is not the same as "in a candidate together". A `.by.` term is one term of a candidate but splits into two for the screen. Measured on the standard fixture at `max.predictors = 1`: an `NA` at `complexity/depth` is accepted, while `depth/ZONE` and `complexity/ZONE` are reported.

This is the one place where the change does not do what the issue text proposed, and it is a decision rather than an oversight — the reasoning is below and in a test comment, so it can be reversed deliberately if you disagree.

## Where the check goes, and why not where the issue said

#27 proposed the check in `build_predictor_correlation_matrix()`'s supplied branch, leaving `combine_uncorrelated()` alone. That was tried first and does not work: `resolve_factor_interactions()` screens factor pairs against the raw supplied matrix and runs *before* that function, so `factor.factor.interactions` still failed with the old message.

A union of every name that might be screened was tried second. That fails in the opposite direction. A name given in the character form of `factor.factor.interactions` is screened against the others in that argument and never against `pred.vars.cont`, so a union treats the Cartesian product as screened and rejects a cell nothing reads. It also still missed a pair crossing a derived `.I.` column and a character-form `smooth.smooth.interactions` name.

**The check now runs at the two screening sites**, on the sub-matrix each is about to take `max(abs(.))` of. That covers exactly the pairs compared, and means no model of which names reach which screen has to be maintained beside the code that does the screening. Both functions are edited only to add that call. Stopping on an `NA` was already the right behaviour at each, the alternative being to drop the combination silently; this makes the message name the pair.

Six paths were verified end to end, five that must fail and one that must not:

| supplied `NA` on | screened? | result |
|---|---|---|
| two factors, `factor.factor.interactions = TRUE` | yes | reported |
| two continuous predictors | yes | reported |
| a derived column supplied in one dimension only | yes | reported |
| a non-predictor named via character `smooth.smooth.interactions` | yes | reported |
| a pair crossing a derived column and such a name | yes | reported |
| a pair no screen reads | no | accepted |
| a computed matrix, no `cor.matrix` argument | — | no longer misattributed to a supplied one |

<details>

### #26

`cor_matrix_supplied()` in `R/generate-model-set.R` replaces the `length(cor.matrix)==1` test. A matrix or data.frame is supplied; a length-one `NA` is the default and is not; anything else is rejected naming its class.

That last part follows from replacing the sentinel rather than being an addition to the issue: something has to happen for the remaining cases, and under the old test a `NULL` or a character value fell through to the supplied branch and failed further on against the missing-predictor check, reporting predictors rather than the argument. The new message is `cor.matrix must be a two-dimensional matrix or data.frame of predictor correlations, or NA to have them computed from use.dat. Supplied: NULL of length 0.`

### #27

Further points, each recorded on the face of the code:

- **Off-diagonal only.** Every screen takes `max(abs(.))` over `upper.tri()` and `lower.tri()`, which exclude the diagonal, so a diagonal `NA` is never read and must not be reported as though it were. There is a test asserting it is accepted.

`combine_uncorrelated()` and `enumerate_candidate_models()` are each edited only to add the call. Stopping on an `NA` was already the right behaviour at both, the alternative being to drop the combination silently; this makes the message name the pair. #27 proposed leaving `combine_uncorrelated()` untouched, which is the part of its proposal this does not follow, for the reason given above.

### Tests

Eleven new `test_that()` blocks in `test-generate_model_set.R`, beside the existing supplied-`cor.matrix` ones: the 1x1 case; four rejected argument classes, two of each length behaviour; an S4 `Matrix` accepted; an `NA` between two predictors; one reaching the factor interaction screen; one in a derived column supplied in one dimension only; one on a name screened without being a predictor; one on a pair crossing a derived column and such a name; and two asserting that an `NA` on an unscreened pair is accepted.

One further test — that an `NA` on the diagonal is accepted — passes before and after. It guards against a false positive the new checks could introduce rather than against the old defect, which is why it is written as `expect_no_error()` and named here rather than presented as a regression test.

Constructing the derived-column test needs two *independently sampled* factors. Two factors derived from one another are correlated above `cov.cutoff`, so no interaction column is built and the cell under test is never read — the first attempt passed for that reason rather than because the code was right.

### Verification

`devtools::check()`: 0 errors, 0 warnings, 0 notes. `testthat::test_local()`: 638 passing, 0 failing, 7 skipped blocks, against 616 at the branch point. Debian WSL2, R 4.6.1, testthat 3.3.2, 2026-09-03.

</details>


beckyfisher:
Eight independent review rounds. Every round from the first to the fifth found a defect; the sixth, seventh and eighth found none in the code. What the reviews changed was not the intent of the fix but where it goes, twice over.

## The check moved three times

**#27 proposed putting the check in `build_predictor_correlation_matrix()`'s supplied branch, leaving `combine_uncorrelated()` alone.** That was the first attempt and it does not work: `resolve_factor_interactions()` screens factor pairs against the raw supplied matrix and runs *before* that function, so `factor.factor.interactions` still failed with `missing value where TRUE/FALSE needed`. A second copy of the `length(cor.matrix)==1` sentinel was found in the same place, so #26 was half fixed too.

**A union of every name that might be screened was the second attempt.** It fails in the opposite direction. A name given in the character form of `factor.factor.interactions` is screened against the others in that argument and never against `pred.vars.cont`, so a union treats the Cartesian product as screened and rejects a cell nothing reads. It also still missed a pair crossing a derived `.I.` column and a character-form `smooth.smooth.interactions` name.

**The check now runs at the two screening sites**, on the sub-matrix each is about to take `max(abs(.))` of. That covers exactly the pairs compared, and no model of which names reach which screen has to be maintained beside the code that does the screening. Six paths were verified, five that must fail and one that must not.

## A defect the change itself introduced

Round 4 found the new error firing on a matrix the package computed itself, while telling the user to fix the one they supplied. `check_correlations()` returns `NA` for a pair involving a single-level factor, and `factor_correlations()` had no zero-fill. So with no `cor.matrix` argument at all, a user was told to "supply a correlation for each" — with nothing to supply. #27 asserts the computed branch "can never reach" the `NA`; that assertion was inherited into the code untested, and is false by this route.

## Two departures from the issues, both decisions

**The report depends on `max.predictors`**, which #27 asked it not to. It covers exactly the pairs some screen compares, so an `NA` on a pair nothing reads is accepted. Rejecting an inert cell would be a false failure of the kind round 3 found. Note that "compared" is not "in a candidate together": a `.by.` term is one term but splits into two for the screen, so at `max.predictors = 1` a continuous predictor and a factor still are compared.

**The zero-fill admits a model set holding the same model twice.** With a single-level factor, `fa.I.const` is `paste(fa, "a")` — identical to `fa` — and both enter the AICc table and the variable importance scores. That is a poor model set and a better outcome than the error it replaces. The real defect is that a single-level factor is accepted as a predictor at all, which is #33.

Both are recorded in `NEWS.md`, in the body and in test comments, so either can be reversed deliberately.

## What the change is bounded by

432 scenarios, regenerated independently in the final round from exports of both trees:

| outcome | cells |
|---|---|
| identical model set | 104 |
| identical error | 208 |
| error before, model set now | 40 |
| error before, a different error now | 80 |
| model set before, a different model set now | 0 |
| model set before, an error now | 0 |

Every one of the 40 cells that now build has `pred.vars.cont = NA`: name a continuous predictor and the call still stops, in `check_correlations()`, on the same single-level factor. No call that produced a model set before produces a different one, or fails, now.

That grid was counted wrongly three times before it was counted right — once comparing error strings rather than model sets, once double-counting a category, once mischaracterising one. `NEWS.md` records the superseded counts so the figures are not taken from them.

## Raised separately

**#37** — the character forms of `smooth.smooth.interactions` and `factor.factor.interactions` are validated against `rownames(cor.matrix)` and `colnames(use.dat)` rather than against the predictor lists, so either accepts a variable that is not a predictor and builds interaction terms over it. Assigned to batch 4.

## State

`devtools::check()`: 0 errors, 0 warnings, 0 notes. `testthat::test_local()`: 638 passing, 0 failing, 7 skipped blocks, against 616 at the branch point. Eleven new `test_that()` blocks; the final round confirmed `man/` and `NAMESPACE` regenerate byte-identically, the snapshot is unchanged, there is no CRLF churn, and all 24 issue references are repository-qualified. Debian WSL2, R 4.6.1, testthat 3.3.2, 2026-09-03.

PR #38, opened 2026-09-03
Batch 4: reject predictor input that produced a model set nothing reported

### LINKED ISSUE #22: A hard coded factor interaction column can collide with an existing column of the same name
## What

`generate_model_set()` appends a hard coded factor interaction column named
`a.I.b` without checking whether `use.dat` already has a column of that name.
`cbind()` accepts the duplicate, and the returned `used.data` carries two columns
with the same name.

## Why it matters

Every later stage selects predictors by name. `use.dat[, "a.I.b"]` returns the
first of the two, which is the user's own column, while the model formulas and
the correlation matrix were built for the generated one. The two are different
variables, and nothing reports the collision.

## Evidence

Two factors named so that their generated interaction column collides with an
existing continuous predictor of the same name, on this repository's
`case_study1`:

```r
use.dat <- FSSgam::case_study1
use.dat$site <- factor(use.dat$site)
use.dat$ZONE.copy <- factor(paste0(use.dat$ZONE, ".copy"))
use.dat$ZONE.I.ZONE.copy <- rnorm(nrow(use.dat))
test.fit <- mgcv::gam(log.Herbivore.biomass ~ s(depth, k = 3, bs = "cr"),
                      data = use.dat)

model.set <- generate_model_set(
  use.dat = use.dat, test.fit = test.fit,
  pred.vars.cont = c("depth", "ZONE.I.ZONE.copy"),
  pred.vars.fact = c("ZONE", "ZONE.copy"),
  factor.factor.interactions = TRUE, max.predictors = 2, k = 3
)
sum(colnames(model.set$used.data) == "ZONE.I.ZONE.copy")
#> 2
```

The name is contrived, but nothing prevents it, and the same shape arises for a
user who runs `generate_model_set()` twice, passing the `used.data` of the first
call as the `use.dat` of the second.

<details>

`build_factor_interaction_columns()` in `R/generate-model-set.R` builds the new
columns and appends them with `cbind()`. The check belongs there, against
`colnames(use.dat)`.

Which behaviour is right is a decision rather than an obvious fix. Stopping with
the colliding name reported is the most conservative. Overwriting the existing
column silently changes a predictor the user named. Warning and overwriting is a
middle course, and matches how the function reports other cases where it declines
to build something.

Found while reviewing FSSgam_package#21; it is pre-existing and independent of
that change.

</details>

### LINKED ISSUE #28: A predictor named twice yields a candidate holding it twice, with -Inf warnings
## What

A predictor named twice in `pred.vars.fact` (or in `pred.vars.cont`) produces a
candidate model holding that predictor twice, such as `ZONE+ZONE`, together with
two "no non-missing arguments to max; returning -Inf" warnings.

## Why it matters

The candidate is not a model anyone asked for, and it is fitted and ranked
alongside the rest. The warnings name nothing that would lead a user to the
duplicated entry.

## Evidence

On `master` at `ed50981`, with `case_study1`:

```r
model.set <- generate_model_set(
  use.dat = case_study1, test.fit = test.fit, pred.vars.cont = "depth",
  pred.vars.fact = c("ZONE", "ZONE"), max.predictors = 2, k = 3
)
names(model.set$mod.formula)
#> "null" "depth" "ZONE" "ZONE+depth" "ZONE+ZONE" "ZONE+depth.by.ZONE"
```

with

```
Warning: no non-missing arguments to max; returning -Inf
```

twice.

<details>

`enumerate_candidate_models()` in `R/generate-model-set.R` deduplicates a
candidate's terms before indexing the correlation matrix
(`terms.m <- unique(mod.terms)`), so a candidate holding one term twice yields a
1x1 sub-matrix, both triangles of which are empty. `max()` of an empty vector
warns and returns `-Inf`, which is below any `cov.cutoff`, so the candidate
survives.

The straightforward fix is to reject a duplicated entry in `pred.vars.cont`,
`pred.vars.fact` or `linear.vars` at the top of `generate_model_set()`, beside
the other argument validation, naming the repeated predictor. That is a clearer
error than anything downstream can produce, and it removes the empty-triangle
case rather than papering over it.

Noticed while reviewing FSSgam_package#21, which changes the neighbouring
`combine_uncorrelated()` so that a duplicated factor no longer yields an
interaction of that factor with itself. The candidate above is unaffected by
that change and remains as described.

</details>

### LINKED ISSUE #33: A single-level factor predictor aborts both correlation functions and generate_model_set() through an unguarded lm()

### LINKED ISSUE #37: smooth.smooth.interactions and factor.factor.interactions accept a variable that is not a predictor
## What

The character forms of `smooth.smooth.interactions` and
`factor.factor.interactions` are each validated against something other than the
predictor list they are supposed to name. `smooth.smooth.interactions` is checked
against `rownames(cor.matrix)`, and `factor.factor.interactions` against
`colnames(use.dat)`. Either will therefore accept a variable that is not a
predictor of the model set, and build interaction terms over it.

## Why it matters

The documentation for both arguments states that the named variables must be
predictors. A user who mistypes a name, or who names a variable they meant to add
to `pred.vars.cont` and did not, gets a model set containing terms in a variable
that appears nowhere else in it — no main effect, no entry in
`included.vars`, and no column in the variable-inclusion table that
`fit_model_set()` returns.

Nothing reports it. The model set is built, fitted and ranked.

## Evidence

Measured on the branch of FSSgam_package#36, Debian WSL2, R 4.6.1, 2026-09-03.

The gap is reachable only when a `cor.matrix` is supplied, because the set the
name is checked against is `rownames(cor.matrix)`. With no matrix supplied, one
is computed over the predictors, so a non-predictor is absent from it and the
existing message fires correctly:

```r
generate_model_set(use.dat = use.dat, test.fit = test.fit, k = 3,
                   pred.vars.cont = c("depth", "complexity"),
                   smooth.smooth.interactions = c("zz", "depth"),
                   max.predictors = 2)
#> Error: Variable(s) zz named in smooth.smooth.interactions are not predictors.
#>        Each must appear in pred.vars.cont (or linear.vars).
```

Supply a matrix that names `zz`, and the same call is accepted:

```r
set.seed(3)
use.dat <- FSSgam::case_study1
use.dat$site <- as.factor(use.dat$site)
use.dat$zz <- rnorm(nrow(use.dat))          # not named as a predictor
test.fit <- mgcv::gam(log.Herbivore.biomass ~ s(depth, k = 3, bs = "cr") +
                      s(site, bs = "re"), data = use.dat)
cm <- check_correlations(use.dat[, c("depth", "complexity", "zz")])

model.set <- generate_model_set(
  use.dat = use.dat, test.fit = test.fit, k = 3,
  pred.vars.cont = c("depth", "complexity"),
  smooth.smooth.interactions = c("zz", "depth"),
  max.predictors = 2, cor.matrix = cm
)

names(model.set$mod.formula)
#> "null" "depth" "complexity" "zz.te.depth"
model.set$included.vars
#> "depth" "complexity"
model.set$mod.formula[["zz.te.depth"]]
#> ~te(zz, depth, k = 3, bs = c("cr", "cr"))
```

A candidate in `zz` is built, fitted and ranked, while `zz` appears in no other
candidate, in `included.vars`, or in the variable-inclusion columns of
`fit_model_set()`'s output.

So the check each argument does have is against the wrong set, and for
`smooth.smooth.interactions` the set depends on an unrelated argument.
`R/generate-model-set.R`: the `smooth.smooth.interactions` character branch
tests against `rownames(cor.matrix)`, and the `factor.factor.interactions`
character branch against `colnames(use.dat)`. Neither is the predictor list.

`factor.factor.interactions` is checked against `colnames(use.dat)`, which does
not depend on a supplied matrix, so any column of `use.dat` is accepted whether
or not one is given.

Found while reviewing FSSgam_package#36, which widened an unrelated check to
cover these names so that a supplied `cor.matrix` with an `NA` between them is
still reported. That change works around this and does not fix it.

<details>

### The check each should make

`smooth.smooth.interactions`, in its character form, names continuous
predictors: every entry should be in `pred.vars.cont`, or in `linear.vars`,
which the existing error message already states as the requirement.

`factor.factor.interactions`, in its character form, names factors: every entry
should be in `pred.vars.fact`.

Both belong beside the other argument validation, and both should name the
offending entry.

### What to settle

Whether this is an error or a warning. Erroring is consistent with how a
misspelled `pred.vars.cont` entry is treated, and with FSSgam_package#28's
proposal for a duplicated predictor. It is a behaviour change: a call naming a
non-predictor currently succeeds.

Note the ordering constraint. `factor.factor.interactions` is consumed by
`resolve_factor_interactions()`, which runs before the resolved correlation
matrix exists, and `smooth.smooth.interactions` by
`resolve_smooth_smooth_interactions()`, which runs after. Validating both at the
top of `generate_model_set()` against the predictor lists avoids depending on
either.

### Definition of done

A variable named in either argument that is not a predictor of the model set is
reported, naming the entry and the argument, rather than silently contributing
interaction terms.

</details>

beckyfisher:
Stacked on #36, which is stacked on #35 and #32. Review those first; this pull request's diff against `dev` includes them until they merge.

## What

Four defects, all of them input `generate_model_set()` accepted and should not have. Each produced a model set that looked ordinary and was not.

Closes #28, closes #37, closes #33, closes #22.

## Why it matters

| issue | accepted input | what the model set then held |
|---|---|---|
| #28 | a predictor named twice, by any of four routes | a candidate `ZONE+ZONE` or `fa.I.fb+fa.I.fb`, and two `-Inf` warnings naming nothing |
| #37 | a non-predictor named in an interaction argument | `te()` terms in a variable with no main effect and no inclusion column |
| #33 | a factor with one level | an interaction column that is a copy of the other factor — the same model twice |
| #22 | a generated column name already in `use.dat` | two columns of one name; formulas built for one, selection returning the other |

None was reported. In every case the run completes, the candidates are fitted, and the AICc table and variable importance scores include them.

## Evidence

**#28**, on `case_study1`:

```r
generate_model_set(..., pred.vars.fact = c("ZONE", "ZONE"), max.predictors = 2)
#> before: names include "ZONE+ZONE", with two warnings:
#>         no non-missing arguments to max; returning -Inf
#> after:  Error: Each predictor may be named once within an argument.
#>         Named twice: ZONE (pred.vars.fact).
```

**#37**, with `zz` a variable deliberately not named as a predictor, and a supplied matrix carrying it. `zz` is drawn uncorrelated: `macro`, the obvious choice from `case_study1`, correlates -0.335 with `depth` and so exceeds the default `cov.cutoff`, which means no `te()` term is built and the reproduction shows nothing.

```r
use.dat$zz <- rnorm(nrow(use.dat))
cm <- check_correlations(use.dat[, c("depth", "complexity", "zz")])
generate_model_set(..., pred.vars.cont = c("depth", "complexity"),
                   smooth.smooth.interactions = c("zz", "depth"), cor.matrix = cm)
#> before: OK — candidates are null, depth, complexity, zz.te.depth,
#>         while included.vars is depth, complexity
#> after:  Error: Variable(s) zz named in smooth.smooth.interactions are not
#>         predictors. Each must appear in pred.vars.cont (or linear.vars).
```

**#33**, a single-level factor:

```r
check_correlations(data.frame(x = rnorm(20), f = factor(rep("a", 20))))
#> before: Error: contrasts can be applied only to factors with 2 or more levels
#> after:  a matrix, with the f/x cell NA

generate_model_set(..., pred.vars.fact = c("fa", "const"))
#> after:  Error: Factor predictor(s) with fewer than two observed levels: const (1).
```

**#22**, a user column of the generated name:

```r
use.dat$fa.I.fb <- rnorm(nrow(use.dat))
generate_model_set(..., pred.vars.fact = c("fa", "fb"), factor.factor.interactions = TRUE)
#> before: two columns named fa.I.fb, silently
#> after:  Warning: Factor interaction column(s) not created, because use.dat
#>         already has a column of the same name: fa.I.fb.
```

## The check that had to be narrowed

The first version of #28's check compared `pred.vars.cont` against `linear.vars` as well. **The existing test suite caught it**: naming a predictor in both is the documented way to fit it linearly, `setdiff(pred.vars.cont, linear.vars)` deciding which predictors get a smooth. The check would have broken a supported idiom.

It now rejects a repeat *within* one argument, and a name given as both a factor and a continuous predictor, and leaves that pair alone.

**The `-Inf` warnings had five further causes, none of which that check reaches.** A name repeated inside `factor.factor.interactions`; two generated interaction names colliding with each other; a predictor whose own name contains a reserved separator, `+`, `*`, or the name `null`, raised as #39 and rejected here; and: The exempt idiom puts the name in the candidate pool twice and hits the same empty triangle at any `max.predictors` of 2 or more. Both screens now test the sub-matrix through `exceeds_cutoff()`, which returns `FALSE` for an empty triangle rather than calling `max()` on it, so no candidate set emits the warning. Which combinations survive is unchanged, `-Inf` having been below every cutoff.

## What this closes from #36

Two things that pull request could only work around are now fixed at source.

Its `NA` check was widened to cover names screened without being predictors; #37 stops such a name reaching a screen at all, so the widening is belt and braces. And zero-filling the computed matrix there turned a single-level-factor error into a model set containing a duplicate model; #33 rejects that factor first, so the duplicate is unreachable.

Measured rather than asserted, since two earlier counts here were wrong: 80 `test_that` blocks across the two modified files at the branch point, 97 now; 5 titles removed and 22 new. Of the removed, all but one are renames whose property is still asserted; the exception is the block that pinned `combine_uncorrelated()`'s by-name indexing, which nothing now pins and which the code records as such.

Two rounds of review found blocks deleted while the body claimed they were rewritten — the same three both times, by a text replacement whose span overreached. They are restored, and a script now compares the list of block names against an earlier ref after every edit to a test file.

<details>

### Where the validation goes

All four checks are in one block at the top of `generate_model_set()`, beside `validate_use_dat()` and `validate_null_terms()`, except #22's, which belongs in `build_factor_interaction_columns()` where the name is generated.

Placement matters for #37. `factor.factor.interactions` is consumed by `resolve_factor_interactions()`, which runs *before* the resolved correlation matrix exists; `smooth.smooth.interactions` by `resolve_smooth_smooth_interactions()`, which runs after. Validating both at the top against the predictor lists avoids depending on either, and on whether a `cor.matrix` was supplied.

### Decisions

**#22 declines rather than overwrites**, and drops the combination from the returned term names as well as the columns, so no later stage refers to a term that was not built. Overwriting would silently change a predictor the user named.

**#33 is two halves.** The `lm()` guard is committed separately, because both correlation functions are exported and callable directly, so it matters independently of `generate_model_set()` rejecting the factor. `droplevels()` runs first, so a level declared but taken by no row counts as absent.

### Commit structure, and a deviation

Seven commits. The first two split the work as far as it splits cleanly: the `lm()` guards are separable by file and are their own commit, while the four validation changes share one block, the same tests and the same `NEWS.md` section, and an attempt to split those produced intermediate commits with duplicated test blocks and an empty `NEWS.md`. The repository's note about folding fixes into one commit applies to that pair, and it is a deliberate exception rather than an oversight.

The remaining five are the review rounds, each closing a further route or restoring an assertion, and each is self-contained.

### Verification

`devtools::check()`: 0 errors, 0 warnings, 0 notes. `testthat::test_local()`: 670 passing, 0 failing, 7 skipped blocks, against 638 at the branch point. No duplicate test names in either modified file.

Note on mutation testing in this repository: `testthat::test_file()` did not pick up an edit to `R/` even after `devtools::load_all()`, and reported no failures for a message that had demonstrably changed. A mutation result here has to be checked against a direct call. Debian WSL2, R 4.6.1, testthat 3.3.2, 2026-09-04.

</details>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -19,10 +19,16 @@
   every cutoff.
 
   A name appearing in both `pred.vars.fact` and `pred.vars.cont` or
-  `linear.vars` is rejected too. A name appearing in both `pred.vars.cont` and
-  `linear.vars` is **not**: that is the documented way to fit a continuous
-  predictor linearly, `setdiff(pred.vars.cont, linear.vars)` deciding which
-  predictors get a smooth.
+  `linear.vars` is rejected too, as is a name repeated within the character form
+  of `factor.factor.interactions` or `smooth.smooth.interactions`. That last is
+  a third route to the same defect and the one the others miss:
+  `factor.factor.interactions = c("fa", "fa", "fb")` built the interaction
+  column `fa.I.fb` twice, so `use.dat` gained two columns of that name and the
+  candidate set gained `fa.I.fb+fa.I.fb`.
+
+  A name appearing in both `pred.vars.cont` and `linear.vars` is **not**
+  rejected: that is the documented way to fit a continuous predictor linearly,
+  `setdiff(pred.vars.cont, linear.vars)` deciding which predictors get a smooth.
 
 * **Behaviour change:** a variable named in the character form of
   `factor.factor.interactions` or `smooth.smooth.interactions` must be a
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,56 @@
 # FSSgam (development version)
 
+* **Behaviour change:** `generate_model_set()` rejects a predictor named twice
+  within `pred.vars.cont`, `pred.vars.fact` or `linear.vars`, naming it. Such a
+  name produced a candidate holding it twice, `ZONE+ZONE`, together with two
+  "no non-missing arguments to max; returning -Inf" warnings naming nothing:
+  `enumerate_candidate_models()` deduplicates a candidate's terms before
+  indexing the correlation matrix, so a candidate holding one term twice gives a
+  1x1 sub-matrix whose triangles are empty, and `max()` of an empty vector
+  returns `-Inf`, which is below any `cov.cutoff` (FSSgam_package#28).
+
+  A name appearing in both `pred.vars.fact` and `pred.vars.cont` or
+  `linear.vars` is rejected too. A name appearing in both `pred.vars.cont` and
+  `linear.vars` is **not**: that is the documented way to fit a continuous
+  predictor linearly, `setdiff(pred.vars.cont, linear.vars)` deciding which
+  predictors get a smooth.
+
+* **Behaviour change:** a variable named in the character form of
+  `factor.factor.interactions` or `smooth.smooth.interactions` must be a
+  predictor of the model set. Each argument was checked against the wrong set --
+  the first against `colnames(use.dat)`, so any column was accepted, and the
+  second against `rownames(cor.matrix)`, so what was accepted depended on
+  whether a `cor.matrix` was supplied and what names it carried. A non-predictor
+  named in either was screened against `cov.cutoff` and contributed interaction
+  terms, while appearing in no other candidate, in `included.vars`, or in the
+  variable-inclusion columns `fit_model_set()` returns (FSSgam_package#37).
+
+* **Behaviour change:** a factor predictor with fewer than two observed levels is
+  rejected, naming it and its level count. Such a factor explains nothing, cannot
+  be screened for collinearity, and produces an interaction column that is a copy
+  of the other factor, so the model set holds the same model under two names.
+  Levels declared but taken by no row are counted as absent (FSSgam_package#33).
+
+* Bug fix: `check_correlations()` and `check_non_linear_correlations()` now wrap
+  their `lm()` call in `try()`, as every other fit in both already was. It was
+  the one unguarded fit in each, so a single-level factor stopped the whole call
+  with "contrasts can be applied only to factors with 2 or more levels" instead
+  of leaving that cell `NA`. Both are exported and callable directly, so the
+  guard matters beyond the `generate_model_set()` path the check above covers
+  (FSSgam_package#33).
+
+* **Behaviour change:** a hard coded factor interaction column whose generated
+  name is already a column of `use.dat` is not created, and a warning names it.
+  `cbind()` accepted the duplicate, and every later stage selects predictors by
+  name: `use.dat[, "a.I.b"]` returned the first of the two, which is the user's
+  own column, while the model formulas and the correlation matrix were built for
+  the generated one. Nothing reported the collision. Declining rather than
+  overwriting, because overwriting silently changes a predictor the user named
+  (FSSgam_package#22).
+
+  This is reached without contrivance: passing the `used.data` of one call as the
+  `use.dat` of the next presents every generated name as an existing column.
+
 * Bug fix: a user-supplied `cor.matrix` with exactly one cell was silently
   discarded and recomputed from `use.dat`. A model set with a single predictor
   is the only case in which a supplied matrix is 1x1, and
--- a/NEWS.md
+++ b/NEWS.md
@@ -72,8 +72,8 @@
   declined and warned about as well, though it is no longer reachable through the
   public API, the predictor name it requires being rejected by the check below.
 
-* **Behaviour change:** a predictor whose name contains `.by.`, `.te.`, `.t.` or
-  `.I.` is rejected, naming it and them. Those strings separate the parts of a
+* **Behaviour change:** a predictor whose name contains `.by.`, `.te.`, `.t.`,
+  `.I.`, `+` or `*` is rejected, as is one named `null`. Those strings separate the parts of a
   generated term name and are found with `grep(fixed = TRUE)` over every term,
   so a predictor named `catch.by.effort` -- an ordinary variable name -- was
   parsed as an interaction. The candidate became
@@ -81,6 +81,15 @@
   predictor vanished from the model set. The only indication was the `-Inf`
   warnings, which this release removes for unrelated reasons
   (FSSgam_package#39).
+
+  `+` joins the terms of a candidate name and `*` writes a linear interaction, so
+  either is parsed as more than one term in the same way. `null` is the name
+  given to the null model's own candidate, so a predictor of that name produced
+  two candidates called `null`: `mod.formula[["null"]]` returned the first, and
+  the variable importance table matched the wrong row.
+
+  `factor.smooth.interactions` in its character form is validated alongside the
+  other two interaction arguments, which it was not.
 
   This rejects the failure rather than fixing the parse. Making a term's
   structure explicit instead of recovering it from its name is a larger change,
--- a/NEWS.md
+++ b/NEWS.md
@@ -69,7 +69,25 @@
   Two generated names can also collide with each other, which is the same defect
   reached from inside rather than outside: `(a, b.I.c)` and `(a.I.b, c)` both
   paste to `a.I.b.I.c`, and the two columns are different variables. That is
-  declined and warned about as well.
+  declined and warned about as well, though it is no longer reachable through the
+  public API, the predictor name it requires being rejected by the check below.
+
+* **Behaviour change:** a predictor whose name contains `.by.`, `.te.`, `.t.` or
+  `.I.` is rejected, naming it and them. Those strings separate the parts of a
+  generated term name and are found with `grep(fixed = TRUE)` over every term,
+  so a predictor named `catch.by.effort` -- an ordinary variable name -- was
+  parsed as an interaction. The candidate became
+  `~s(catch.by.effort) + s(catch, by = effort)`, which cannot be fitted, so the
+  predictor vanished from the model set. The only indication was the `-Inf`
+  warnings, which this release removes for unrelated reasons
+  (FSSgam_package#39).
+
+  This rejects the failure rather than fixing the parse. Making a term's
+  structure explicit instead of recovering it from its name is a larger change,
+  and the issue records what it would involve. One consequence is worth knowing:
+  passing the `used.data` of one call as the `use.dat` of the next presents
+  generated names, which contain `.I.` by construction, so those columns must be
+  dropped or renamed before they can be named as predictors.
 
 * Bug fix: a user-supplied `cor.matrix` with exactly one cell was silently
   discarded and recomputed from `use.dat`. A model set with a single predictor
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,15 @@
   1x1 sub-matrix whose triangles are empty, and `max()` of an empty vector
   returns `-Inf`, which is below any `cov.cutoff` (FSSgam_package#28).
 
+  The warnings had a second cause, which rejecting the repeated name does not
+  address: naming a predictor in both `pred.vars.cont` and `linear.vars`, the
+  documented idiom below, puts the name in the candidate pool twice and reaches
+  the same empty triangle at any `max.predictors` of 2 or more. Both screens now
+  test the sub-matrix through a helper that returns `FALSE` for an empty
+  triangle rather than calling `max()` on it, so no candidate set emits the
+  warning. Which combinations survive is unchanged, `-Inf` having been below
+  every cutoff.
+
   A name appearing in both `pred.vars.fact` and `pred.vars.cont` or
   `linear.vars` is rejected too. A name appearing in both `pred.vars.cont` and
   `linear.vars` is **not**: that is the documented way to fit a continuous
--- a/NEWS.md
+++ b/NEWS.md
@@ -66,6 +66,11 @@
   This is reached without contrivance: passing the `used.data` of one call as the
   `use.dat` of the next presents every generated name as an existing column.
 
+  Two generated names can also collide with each other, which is the same defect
+  reached from inside rather than outside: `(a, b.I.c)` and `(a.I.b, c)` both
+  paste to `a.I.b.I.c`, and the two columns are different variables. That is
+  declined and warned about as well.
+
 * Bug fix: a user-supplied `cor.matrix` with exactly one cell was silently
   discarded and recomputed from `use.dat`. A model set with a single predictor
   is the only case in which a supplied matrix is 1x1, and
--- a/R/check-correlations.R
+++ b/R/check-correlations.R
@@ -74,12 +74,18 @@ build_continuous_correlation_matrix=function(dat,cont.vars){
 build_factor_continuous_skeleton=function(dat,fact.vars,cont.vars,cor.mat){
    if(length(cont.vars)>0){
     lm.grid=expand.grid(list(fact.var=fact.vars,cont.var=cont.vars))
-    # The lm() below is not wrapped in try(), unlike the multinom() fits in
-    # fill_factor_factor_correlations(). A single-level factor makes it stop and
-    # the error propagates out of check_correlations(): FSSgam_package#33. The
-    # same shape is present in check_non_linear_correlations().
+    # try(), as the multinom() fits in fill_factor_factor_correlations() have.
+    # A single-level factor makes lm() stop with "contrasts can be applied only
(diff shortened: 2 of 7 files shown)


beckyfisher:
Six independent review rounds. Every round from the first to the fifth found both a further route by which a predictor was silently duplicated or dropped, and an assertion this branch had lost or weakened. The sixth found neither.

## What the branch rejects, and why each mattered

Four issues were in scope. Reviewing them found three more routes to the same class of defect, all now closed.

| input | what the model set held, unreported |
|---|---|
| a predictor named twice in one argument | a candidate `ZONE+ZONE` |
| a name given as both a factor and a continuous predictor | a term written both ways |
| a name repeated inside `factor.factor.interactions` | `fa.I.fb` built twice — two `used.data` columns, and `fa.I.fb+fa.I.fb` |
| two combinations pasting to one name | two `used.data` columns holding different variables |
| a non-predictor named in an interaction argument | `te()` terms in a variable with no main effect and no inclusion column |
| a predictor whose name contains `.by.`, `.te.`, `.t.`, `.I.`, `+` or `*` | nothing — the predictor vanished from the model set |
| a predictor named `null` | two candidates named `null`; the importance table matched the wrong row |
| a factor with one observed level | an interaction column that is a copy of the other factor |
| a generated column name already in `use.dat` | two columns of one name; formulas built for one, selection returning the other |

The last three of the routes are #39, filed during review, along with a seventh — `-` and `:` mis-parse the same way — which is recorded there rather than closed here, being less silent: the candidate is retained and the failure appears in `failed.models`.

## The mistake worth recording

**Suppressing the `-Inf` warnings hid three defects that had not been closed.** Those warnings named nothing and fired on a legitimate input, so removing them was right in itself. But they were the only evidence of the third, fourth and fifth routes above, and each round found the next one only because the round before had asked what else the suppression had hidden.

The rule: silence a warning only once every defect it was the sole evidence of is closed.

## Assertions, and how they were lost

Five consecutive rounds found a lost or weakened assertion.

Twice the same three test blocks were deleted while three documents said they had been rewritten — the second time by the commit written to restore them, both times by a text replacement whose span overreached. A script comparing `test_that` names now runs after every edit to a test file.

**That check catches deletion, not weakening**, and the last two losses were weakenings it could not see: a restored block that could not fail, and a branch whose covering test had been rewritten into one that did not discriminate. Both were found only by mutation.

So every claim this branch makes is now mutation-tested — the code a claim depends on is changed, and a test must fail. All are pinned except `combine_uncorrelated()`'s by-name indexing and the self-collision screen, and rather than propping those up with contrived tests, the code and the tests record that nothing pins them.

Twice a path was recorded as unreachable without an attempt to reach it, and both times that was wrong.

## Verification

`devtools::check()`: 0 errors, 0 warnings, 0 notes. `testthat::test_local()`: 670 passing, 0 failing, 7 skipped blocks, against 638 at the branch point. 80 `test_that` blocks across the two modified files at the branch point and 97 now; 5 titles removed, 22 new.

The final round re-ran the whole mutation table independently on a clean `git archive` copy, one fresh process per mutation, and reproduced every row; mutated eight further things, all pinned; and confirmed no reserved string occurs in any of the four bundled datasets, 133 columns, or in any predictor vector in the companion repository's vignettes, so nothing legitimate is refused.

Debian WSL2, R 4.6.1, testthat 3.3.2, 2026-09-04.

Issue #34, opened 2026-09-03
An mgcv::gamm() test.fit cannot be used, and the documentation points users at it


beckyfisher:
## What

An `mgcv::gamm()` fit cannot be used as `test.fit`. `generate_model_set()` stops
on the null model before any candidate is enumerated. Several parts of the
package's documentation describe `gamm` as supported.

## Why it matters

`generate_model_set()`'s own documentation points the user at `gamm`. The
`null.terms` parameter text states:

> to fit a correlation structure only (but no random effects) this must be
> achieved through a call the gamm

and the `@details` of the same function explains at length that `gamm` is based on
`nlme` and so allows correlation structures, which `gam` does not. `r2.type` in
`fit_model_set()` describes its values as what is reported "by gam, gamm or
gamm4", and `extract_mod_dat()` has an `inherits(mod.fit, "gamm")` branch for
`r2.type = "r2"`. A user following that advice reaches an error, and the error
tells them to do something they were not doing.

A correlation structure is the reason someone reaches for `gamm` — repeated
measures on a site, a time series — so this is not an obscure corner. The
practical position is that `uGamm` is the only supported route to a mixed fit, and
the documentation does not say so.

## Evidence

Measured on `dev` at `656ef62`, Debian WSL2, R 4.6.1, mgcv 1.9-4.

```r
use.dat <- FSSgam::case_study1
use.dat$site <- as.factor(use.dat$site)
test.fit <- mgcv::gamm(Herbivore.abundance ~ s(depth, k = 3, bs = "cr"),
                       random = list(site = ~1), data = use.dat)

class(test.fit)        #> "gamm" "list"
is.null(test.fit$call) #> TRUE

generate_model_set(use.dat = use.dat, test.fit = test.fit,
                   pred.vars.cont = c("complexity", "depth"),
                   max.predictors = 2, k = 3)
#> Error: Null model not successfully fitted, please check your inputs.
#>   If there are no random effects try using 'gam' instead of 'uGamm'
#>   in your test.fit model call.
#>   The following error message was provided:
#>   Error in update(test.fit, formula = null.formula, data = use.dat) :
#>     need an object with call component
```

The cause is that `mgcv::gamm()` returns a two-element list of a `lme` and a `gam`
and records no `call` of its own, so `stats::update()` has nothing to re-evaluate.
`MuMIn::uGamm()` exists in part to supply that, which is why `uGamm` works and
bare `gamm` does not.

[...]

Issue #23, opened 2026-09-01
Predictors correlated with a null.terms variable are never screened by cov.cutoff


beckyfisher:
## What

`cov.cutoff` screens correlations among `pred.vars.cont`, `pred.vars.fact` and
`linear.vars` only. Terms supplied through `null.terms` are never part of that
matrix, so a candidate predictor may be arbitrarily strongly correlated with a
`null.terms` variable and will still be admitted to every model in the set.

## Why it matters

`null.terms` is the only mechanism for forcing a term into every candidate, so it
is what a user reaches for when a variable must be present on external grounds —
a random effect, or a confounder that a causal argument requires be adjusted for.
Having done that, the user has good reason to believe collinearity is handled:
`cov.cutoff` defaults to 0.28, which is strict enough that pairs among the
candidates are excluded readily, and the documentation describes the function as
automatically checking for collinearity.

The consequence is silent. A candidate correlated 0.9 with a forced term inflates
the variance of that forced term's estimate, which is frequently the term the
analysis exists to estimate, and nothing in the output indicates it. The model set
looks well screened because every pair it *did* screen passed.

This is not an argument for putting `null.terms` inside the screen. Excluding a
forced term is exactly what must not happen. The gap is that the asymmetry is
neither documented nor reported.

## Evidence

`R/generate-model-set.R`. The screened set is built at line 117 and does not
include `null.terms`:

```r
all.predictors = unique(stats::na.omit(c(pred.vars.cont, pred.vars.fact, linear.vars)))
```

`combine_uncorrelated()` then drops only combinations whose *internal* pairwise
correlations exceed the cutoff:

```r
cor.mat.m = cor.matrix[row.index, col.index]
if (max(abs(cor.mat.m[upper.tri(cor.mat.m)])) > cov.cutoff) { out = NA }
```

`null.terms` reaches `build_model_formulas()` separately and is pasted verbatim
onto every candidate formula:

```r
formula.m = stats::as.formula(paste("~", paste(c(all.terms.vec, null.terms), collapse = "+")))
```

so it is present in every fitted model and absent from every correlation check.

<details>
<summary>Implementation detail</summary>

### Options

**Document only.** Add to the `null.terms` and `cov.cutoff` parameter
documentation that terms entering through `null.terms` are outside the
collinearity screen, and that the user is responsible for checking candidates
against them. Cheapest, and sufficient if the behaviour is considered acceptable.

[...]

PR #40, opened 2026-09-04
Batch 5: screen predictors against forced terms, and reject a bare gamm test.fit

beckyfisher:
Stacked on beckyfisher/FSSgam_package#38, which is stacked on beckyfisher/FSSgam_package#36, beckyfisher/FSSgam_package#35 and beckyfisher/FSSgam_package#32. Review those first; this pull request's diff against `dev` includes them until they merge.

## What

Two issues about terms the user forces into every model, and about the fit those models are built from.

Closes beckyfisher/FSSgam_package#23, closes beckyfisher/FSSgam_package#34.

beckyfisher/FSSgam#18, originally planned for this batch, is deferred at the maintainer's direction: it adds a feature rather than fixing a defect, and its design decisions are better reviewed on their own.

## Why it matters

**#23.** `null.terms` is the only way to force a term into every candidate, so it is what a user reaches for when a variable must be present on external grounds — a confounder a causal argument requires be adjusted for. Having done that, the user has reason to believe collinearity is handled: `cov.cutoff` defaults to 0.28, strict enough that candidate pairs are excluded readily, and the documentation describes the function as automatically checking for collinearity.

But `cov.cutoff`'s matrix covers `pred.vars.cont`, `pred.vars.fact` and `linear.vars` only. A candidate could be correlated 0.99 with a forced term and appear in every model in the set. The consequence is silent: it inflates the variance of the forced term's estimate, frequently the term the analysis exists to estimate, and the model set looks well screened because every pair it *did* screen passed.

**#34.** A `test.fit` produced by `mgcv::gamm()` cannot be used at all — it records no call, so `stats::update()` has nothing to re-evaluate. The error named neither the argument nor the remedy, and the message that followed advised the user to stop using `uGamm`, which is what they would have had to use to succeed. Meanwhile `generate_model_set()`'s own documentation directed them to `gamm` for a correlation structure.

## Evidence

**#23**, with a predictor built to be near-identical to a forced term:

```r
use.dat$depth.copy <- use.dat$depth + rnorm(nrow(use.dat), 0, 0.01)
generate_model_set(..., pred.vars.cont = c("complexity", "depth.copy"),
                   null.terms = "s(depth,k=3,bs='cr')+s(site,bs='re')")
#> before: depth.copy appears in the model set, correlated 1.00 with the
#>         forced depth term, unreported
#> after:  Warning: Predictor(s) dropped, correlated with a null.terms variable
#>         above null.cov.cutoff (0.8): depth.copy (depth, max 1). ...
```

**#34**:

```r
test.fit <- mgcv::gamm(Herbivore.abundance ~ s(depth, k = 3, bs = "cr"),
                       random = list(site = ~1), data = use.dat)
is.null(test.fit$call)   #> TRUE
generate_model_set(use.dat = use.dat, test.fit = test.fit, ...)
#> before: Error: Null model not successfully fitted ... If there are no random
#>         effects try using 'gam' instead of 'uGamm' in your test.fit model call.
#>         [the user did not call uGamm]
#> after:  Error: A test.fit produced by mgcv::gamm() cannot be used: it records
#>         no call ... Fit the same model through MuMIn::uGamm(), which supplies
#>         the call -- for a correlation structure, uGamm(..., correlation = ...)
```

## The decision the test suite forced

**Both directions are read, and a forced term named in either dimension of a supplied matrix is used.** Three orientations of one asymmetry were found in successive reviews — the value read, the row shape and the column shape — each time because the code was written for the square, symmetric case and the general one assumed to follow.

 `check_non_linear_correlations()` returns an asymmetric matrix — row is the response, column the predictor — and only `[candidate, forced]` is large when a candidate is a deterministic function of a forced term. Measured on a squared forced term in the committed test: 0.280 one way — below the default cutoff — and 0.991 the other. `exceeds_cutoff()` has read both triangles for `cov.cutoff` all along.

**A variable inside a `bs='re'` smooth is excluded from the screen.** This was not planned. Two existing tests failed on the first implementation, and both were right to.

The companion repository's case study 2 uses `null.terms = "s(Location,Site,bs='re')"` with `Status` nested in `Location` — correlation 1. A second fixture builds a factor from `site` while `site` is the random effect. **A random-effect grouping factor is correlated with the predictors measured within it by construction.** That is the design of the study, not collinearity to screen out, and dropping `Status` there would be wrong.

So the screen applies to fixed forced terms — the case beckyfisher/FSSgam_package#23 actually describes — and the issue's own note that a grouping factor is "a legitimate and common specification" is now what the code does.

## Behaviour changes

Three, all in `NEWS.md`.

A predictor correlated with a fixed forced term above `null.cov.cutoff` is dropped. Set the argument to 1 to admit every predictor whatever its correlation. It is also validated: a single non-negative number, where an `NA` previously gave an internal error and a length-2 value was accepted silently.

A bare `mgcv::gamm()` `test.fit` is rejected. It never worked; the change is that the message now says so and names the remedy.

Where dropping leaves fewer predictors than `max.predictors`, the error names the drop rather than reporting only that `max.predictors` is too large.

<details>

### beckyfisher/FSSgam_package#23

`null_term_variables()` recovers the variables by parsing `null.terms`, which is a formula fragment rather than a variable list, and excludes any term setting `bs` to `"re"`. The exemption is read from the parsed call, not by matching the term text: a regex missed `bs=c('re')`, which mgcv accepts and which fits the identical null model, so a nested factor was dropped and the call stopped with a message naming `max.predictors` — none of it the cause. A name that is not a column of `use.dat` — a function, or a term written over several columns — is skipped, a correlation not being defined for it.

`build_null_term_correlations()` uses a supplied `cor.matrix` for any forced term it names in either dimension, and computes the rest of the block from `use.dat`.

The first version used *only* what a supplied matrix names, and that was wrong: such a matrix is indexed by predictor and a forced term is not one, so in the ordinary case nothing was screened and no warning was raised — reinstating for every `cor.matrix` caller exactly the silence beckyfisher/FSSgam_package#23 exists to end.

beckyfisher/FSSgam_package#13 does not force that reading. Its guarantee is that a supplied matrix replaces the automatic estimate *for the predictors*, so a predictor of a class `check_correlations()` cannot classify can be used. Only the `null.vars` block is computed, and where that computation fails — which is what such a predictor causes — the screen is skipped with a warning rather than the call aborting. That caller keeps their model set and is told their forced terms were not screened.

A defect found in the process: `cor.matrix` is reassigned to the *resolved* matrix before the screen runs, so testing that variable for "did the caller supply one" was always true. Whether one was supplied is now recorded once, at the top.

`screen_against_null_terms()` drops from every list a term can reach the candidate set through, and drops an interaction term when any of its parts is dropped, since the term cannot be built without them.

### beckyfisher/FSSgam_package#34

The check is in `build_null_model()`, before `update()` is attempted, and tests both `inherits(test.fit, "gamm")` and `is.null(test.fit$call)` — `gamm4` fits are a different class and unaffected, and `MuMIn::uGamm()` supplies the call that bare `gamm()` lacks.

There is a test that the remedy the message names actually works, not only that the message appears.

Three documentation passages are corrected: `generate_model_set()`'s `null.terms` text, which directed the user to `gamm`; its `@details`, which contrasts `gam` and `gamm` at length; and `fit_model_set()`'s `r2.type`, which names `gamm` among the classes it reports for.

### Commit structure

Nine commits. The first covers both issues together: they share `R/generate-model-set.R`, the same test file and the same `NEWS.md` section, and two earlier attempts in this work to split such a commit by scripted surgery produced intermediate states with duplicated test blocks and an empty `NEWS.md`. That pairing is a deliberate exception to the repository's one-commit-per-bug note rather than an oversight. The remaining six are the review rounds, each self-contained.

### Verification

`devtools::check()`: 0 errors, 0 warnings, 0 notes. `testthat::test_local()`: 772 passing, 0 failing, 7 skipped blocks, against 670 at the branch point.

Assertions here were added in three passes, because two statements about what they covered turned out to be wrong. Thirteen claims were pinned by nothing when first written; the statement that the first five had been fixed was wrong for two of them — `full_subsets_gam()`'s forwarding and three of four separators in the interaction-term drop both left the suite green, the latter allowing two `te()` terms built on a dropped predictor back into the model set.

The mechanism this pull request exists to introduce was also unpinned: nothing tested that a supplied `cor.matrix` *overrides* the computed estimate, because the test set its supplied cell above the cutoff and so passed either way; and the `try()` fallback and its warning — the whole beckyfisher/FSSgam_package#13 argument — had no test at all. A third pass found six more surviving mutations, and a fourth found that the statement fixing them was itself wrong for one — the zero-fill of the *computed* block, reachable with a zero-variance predictor. Each count quoted is read from a run. The claim that all reachable mutations were pinned has itself been wrong twice, so it is not made here: what is claimed is that every mutation named in this pull request was run and its count recorded.

Five statements about coverage on this branch turned out to be false, and the fifth round then found the defect they had been hiding: the screen read only one direction of an asymmetric correlation estimate, so a candidate that is a deterministic function of a forced term entered every model with nothing reported. It survived four rounds because nothing exercised `non.linear.correlations = TRUE` at all — replacing that branch with `if(FALSE)` left the suite green.

The pattern is worth naming. Coverage was repeatedly asserted from having written a test rather than from having run the mutation, and an argument with no test is not covered by the tests of the argument beside it. Debian WSL2, R 4.6.1, testthat 3.3.2, 2026-09-04.

</details>


beckyfisher:
Eight independent review rounds. Four found code defects; four found no defect and instead found statements of mine about test coverage that were untrue. The two are related, and that relation is the useful thing this pull request records.

## One mistake, made four times

Every code defect found after the first round was the same: **code written for the square, symmetric, fully-supplied case, with the general case assumed to follow.**

| round | what was assumed | what was true |
|---|---|---|
| 5 | the correlation estimate is symmetric | `check_non_linear_correlations()` returns row = response, column = predictor, and only one direction is large when a candidate is a function of a forced term |
| 6 | both directions exist and neither is `NA` | `pmax()` with `na.rm = FALSE` made an `NA` in one direction discard a real value in the other |
| 6 | a supplied matrix names each term in both dimensions | a matrix naming a forced term in rows only aborted with `subscript out of bounds` |
| 7 | a forced term is found in the row names | a matrix naming it in columns only was discarded entirely, and a computed estimate substituted |

Measured for the first of those: on a squared forced term, `[forced, candidate]` is 0.280 — below the default cutoff — while `[candidate, forced]` is 0.991. The candidate entered every model unreported, which is the condition beckyfisher/FSSgam_package#23 exists to end.

`exceeds_cutoff()` had read both triangles for `cov.cutoff` all along. The forced-term screen now does the same, and `from.supplied` is allocated over the union of both dimnames rather than subset from either.

## Why four rounds did not find them

**Nothing exercised `non.linear.correlations = TRUE`.** Replacing that branch with `if(FALSE)` left the entire suite green. The asymmetric path had no test, so no claim about coverage could be checked against it — which is why the coverage statements were wrong at the same time.

The lesson generalises: *an argument with no test is not covered by the tests of the argument beside it*, and *a statement that something is mutation-tested is itself a claim*. That claim was wrong five times here before it was made by running the mutation and reading the count.

Every assertion this pull request adds has now been mutated. Of 35 mutations on the final version of the changed code, 32 fail the suite; the three survivors are named below rather than described as covered.

## What the change is

**#23.** `null.cov.cutoff`, defaulting to 0.8, and a returned `null.term.correlations`. `null.terms` forces a term into every candidate, and those terms were outside `cov.cutoff`'s matrix, so a candidate could be correlated 0.99 with a forced term and appear in every model in the set, inflating the variance of the estimate the analysis exists to produce.

**A variable inside a `bs='re'` smooth is exempt**, and the existing suite forced that: the companion repository's case study 2 uses `s(Location,Site,bs='re')` with `Status` nested in `Location`, correlation 1. A grouping factor is correlated with the predictors measured within it by construction — the design of the study, not collinearity to screen out.

**#34.** A bare `mgcv::gamm()` fit records no call, so `stats::update()` has nothing to re-evaluate and every candidate refit fails. It is rejected before the attempt, naming `MuMIn::uGamm` as the remedy, and a test asserts the remedy works rather than only that the message appears. Three documentation passages that directed the user to `gamm` are corrected.

## Left open, deliberately

**#41**, raised at the last round: an `NA` in *both* supplied directions between a forced term and a predictor is read as zero, so the pair is not screened and nothing is reported — while an `NA` between two predictors stops the call and names it, which is beckyfisher/FSSgam_package#27. Either reading is defensible; having one screen stop and the other silently admit is not. Deciding it at the eighth review cycle would be a behaviour change made without review, so the issue records the three options and the constraint that the per-direction fill must stay.

Three surviving mutations are recorded as unpinned rather than described as covered: `combine_uncorrelated()`'s by-name indexing and the self-collision screen, both defensive given their callers, and one path unreachable through the public function.

**#25** is not here. It adds a feature rather than fixing a defect, and was split out at your direction.

## State

`devtools::check()`: 0 errors, 0 warnings, 0 notes. `testthat::test_local()`: 772 passing, 0 failing, 7 skipped blocks, against 670 at the branch point. The final round compared `build_null_term_correlations()` against an independently written expectation over 16 constructed matrix shapes and 2,420 randomised ones, with no value or dimname mismatch. Debian WSL2, R 4.6.1, testthat 3.3.2, 2026-09-04.

PR #46, opened 2026-09-05
Compute a forced-term pair the supplied cor.matrix leaves NA in both directions

beckyfisher:
## What

A pair a supplied `cor.matrix` gives no value for, between a variable named in
`null.terms` and a predictor, is now computed from `use.dat` instead of being
read as a correlation of zero. That is `NA` in both directions, or `NA` in the
one direction a matrix naming the forced term in a single dimension has. This is
the option ruled for on the issue.

## Why it matters

Read as zero, the pair was not screened against `null.cov.cutoff`: the predictor
was admitted to every candidate alongside the forced term, and nothing was
reported. That is the outcome the screen exists to prevent, decided by a cell the
user left empty rather than by a correlation. The same `NA` between two
predictors stops the call and names the pair (FSSgam_package#27), so one screen
stopped and the other silently admitted.

## Evidence

The issue's own reproduction, run on this branch (Debian WSL2, R 4.6.1):

```
Warning: Predictor(s) dropped, correlated with a null.terms variable above
null.cov.cutoff (0.8): a (depth, max 1).
      a complexity
depth 1          0
included: complexity
```

`depth`/`a` is computed at 1.00 and `a` is dropped. The supplied zero at
`depth`/`complexity` is retained, where the computed estimate is 0.334 -- so the
merge is cell by cell, not a recomputation of the row.

The branch was compared against `dev` at 023cd1a with a 20-scenario harness over
`build_null_term_correlations()` called directly, the two results compared with
`identical()`. Four scenarios differ -- `NA` in both directions, `NA` in the one
direction a column-only matrix has, a part-supplied row, and `NaN` -- and every
other scenario is unchanged, including the three failure paths, both
`non.linear.correlations` settings and a supplied `data.frame`.

Three tests added, all confirmed to fail against the unmodified function and to
pass with it. Measured with `testthat::test_local()` and
`sum(as.data.frame(res)$passed)` on this host (R 4.6.1, testthat 3.3.2): 789
passing, 0 failing, 7 skipped, against 772 passing at the branch point.
`devtools::check()` reports 0 errors, 0 warnings, 0 notes.

<details>

### The change

`build_null_term_correlations()` (`R/generate-model-set.R`) decided by row. The
supplied block was allocated over the forced terms the matrix named in either
dimension, every cell no supplied direction covered left at zero, and
`to.compute` was the forced terms the matrix did not name at all. A forced term
the matrix named was therefore never computed, whatever its row left empty.

It now decides by cell. Two matrices are allocated over every forced term and
every screened predictor: the values, and a logical `given` recording which cells
a supplied direction actually gave a value for. `given` is set from `!is.na(blk)`
inside `fill()`, before that direction is zero-filled -- the ordering constraint
the issue named, since after `pmax()` combines the two directions an unsupplied
cell and a supplied zero are the same number. It is set from the transposed block
as well, and after the transposition, so a matrix naming the forced term only as
a column is read the same way. `to.compute` is then every forced term with any
unsupplied cell, and the computed block is merged in at those cells only:

```r
to.compute=null.vars[apply(!given,1,any)]
...
block=out[to.compute,screened,drop=FALSE]
need=!given[to.compute,screened,drop=FALSE]
block[need]=computed[need]
out[to.compute,screened]=block
```

### Behaviour that is unchanged

Where one direction holds a value, that value is used and the cell is not
computed. A matrix that gives a value for every pair takes the early return, as
before, and the no-`cor.matrix` path computes the whole block, as before.

Where the computation fails, the return is unchanged as well: only the rows a
supplied direction gave at least one value for, and `NULL` where none did. A row
of zeros there would state a correlation for the pairs the warning has just said
were not screened, which is the reading this change removes from the supplied
path.

```r
if(!any(has.supplied)){return(NULL)}
return(out[has.supplied,,drop=FALSE])
```

### Two consequences

The computation an unsupplied cell now triggers can fail on a predictor of a
class `check_correlations()` cannot classify, so the existing skip-with-warning
path (FSSgam_package#13) is reached by a route that did not exist. The warning
distinguishes the two cases: it names the pairs where part of the row was
supplied, and the terms where none of it was. The existing wording is kept for
the second, which is what the two committed tests of that path assert.

The row order of `null.term.correlations` is now `null.vars` order rather than
supplied rows followed by computed rows. No test asserted the old order.

### Documentation

`null.cov.cutoff` and `cor.matrix` both state the rule per pair rather than per
forced term, and `@return` records the two cases in which a row or the whole
matrix is absent. The `NEWS.md` `null.cov.cutoff` entry gains a paragraph. The
existing test titled "an NA in the supplied forced-term block is treated as zero"
is retitled: it supplies `NA` in one direction only, and now passes because the
reverse direction supplies zero.

</details>

Closes FSSgam_package#41

https://claude.ai/code/session_018TFFEtmLa4p5oihQ9UQw1A


beckyfisher:
Reviewed over five cycles, each a read of the diff followed by a measurement.
One commit follows, `2de3c65`, and the body has been updated to the current
state -- the Evidence section now reports the comparison against `dev` and the
test count, and the implementation block records the failure-path return and the
transposed-direction bookkeeping.

**Verification.** A 20-scenario harness over `build_null_term_correlations()`
called directly, `dev` at 023cd1a against this branch, compared with
`identical()`. This follows the golden-master approach of Phase 6/6b and Phase
14, the committed suite not reaching every orientation of a supplied matrix.
Scenarios: no matrix; all-supplied zeros; `NA` in both directions; `NA` in one
direction; the forced term named only as a row and only as a column, with and
without `NA`; a part-supplied row; `NaN`; asymmetric supplied values; no forced
terms; every predictor a forced term; a failed computation with nothing
supplied, with part of the block supplied and with no matrix at all; and each of
`non.linear.correlations = TRUE`, a `data.frame` matrix and a factor predictor.
Four scenarios differ, and they are the intended ones. Every other scenario is
`identical()` to `dev`.

**Behaviour.** The failure path returns what it returned on `dev`: only the rows
a supplied direction gave at least one value for, and `NULL` where none did. A
row of zeros there would state a correlation of zero for exactly the pairs the
warning has just said were not screened, which is the reading this change removes
from the supplied path. Two expectations pin it -- `expect_null()` on the wholly
failed path, and that a forced term neither supplied nor computed has no row.

**Scope.** The rule is a pair the matrix gives no value for in any direction it
has, which is wider than `NA` in both directions: a matrix naming the forced term
in a single dimension has one direction, and an `NA` there is the same absence.
`NEWS.md`, both `@param` entries and the internal comments state it per pair
rather than per forced term, the merge being cell-wise, and `@return` records the
two cases in which a row or the whole matrix is absent. A test covers the
column-only shape, in one loop over the two matrix shapes, following the `shapes`
loop the neighbouring test for the row and column orientations already uses.

**Measurements.** This host, Debian WSL2, R 4.6.1, testthat 3.3.2, measured with
`testthat::test_local()` and `sum(as.data.frame(res)$passed)`: 789 passing, 0
failing, 7 skipped, against 772 at the branch point. `devtools::check()` reports
0 errors, 0 warnings, 0 notes. `man/` is in sync with roxygen.

https://claude.ai/code/session_012ubpoYJ8nQcKAQqR7dt9RT

Issue #44, opened 2026-09-05
AICc is NA for every candidate of a quasi-likelihood test.fit, so the model set cannot be ranked


beckyfisher:
## What

`fit_model_set()` returns `mod.data.out$AICc` as `NA` for every candidate of a
quasi-likelihood `test.fit`, so the model set cannot be ranked. `MuMIn::AICc()`,
which `extract_mod_dat()` reads the criterion from
(`R/functions_supporting.R`), returns `NA` because a quasi-likelihood has no
log-likelihood and so no AIC.

## Why it matters

Ranking the candidate set by AICc is what this package does. With the column
entirely `NA`, `delta.AICc` and the Akaike weights are `NA` as well, and
`compute_variable_importance()` has nothing to weight, so the whole output of
the call is unusable.

Nothing reports it. The fits succeed, every candidate is recorded as fitted, and
the failure is visible only on reading the returned table.

`quasipoisson()` and `quasibinomial()` are ordinary choices for overdispersed
count and proportion data, which is most of what this package is used on, so
this is not an obscure corner. `dsm::dsm()` defaults to `quasipoisson()`, which
is how it was found.

## Evidence

Measured on Debian WSL2, R 4.6.1, mgcv 1.9-4, dsm 2.3.4, MuMIn 1.48.19,
2026-09-04, while investigating FSSgam_package#30.

A `dsm` test.fit built from the `dsm` package's own `mexdolphins` data with
`family = quasipoisson()`, and a three-candidate model set fitted from it: every
candidate fits, and `mod.data.out$AICc` is `NA` throughout. It affects
`parallel = FALSE` identically, so it is unrelated to the parallel path this was
found on.

The mechanism is `stats::logLik()` on a quasi-likelihood fit, which mgcv returns
as `NA`. It is a property of the family rather than of this package.

<details>
<summary>Implementation detail</summary>

### Where

`extract_mod_dat()` in `R/functions_supporting.R` reads `MuMIn::AICc(mod.fit)`
and `stats::BIC(mod.fit)`. Neither is defined for a quasi-likelihood.

### Options

**Reject the family.** Refuse a quasi-likelihood `test.fit` in
`generate_model_set()`, beside the `mgcv::gamm()` rejection added for
FSSgam_package#34, naming the family and saying that the model set cannot be
ranked. The most direct, and consistent with how an unusable `test.fit` is now
handled.

**Report it.** Fit as now, and warn once from `fit_model_set()` where the
[...]

Issue #42, opened 2026-09-05
AICc and BIC under mgcv's censored families are not built from a log-likelihood, so the top-model set, the weights and variable importance are wrong


beckyfisher:
## What

`extract_mod_dat()` reads the criterion from `MuMIn::AICc(mod.fit)` and
`stats::BIC(mod.fit)` (`R/functions_supporting.R:59-60`). Both resolve to
`mgcv:::logLik.gam()`, which reads the fitted family's own `aic` slot. For
`mgcv`'s two censored continuous families, `cnorm()` and `clog()`, that slot
does not return a censored log-likelihood.

The fitted models are sound. `dev.resids` and `ls` contain none of the defects,
so REML smoothing-parameter selection, the coefficients and the scale are
unaffected. Only the reported criterion is wrong — and with it every FSSgam
output derived from it: `delta.AICc`, `delta.BIC`, `wi.AICc`, `wi.BIC`, the
top-model set, and variable importance.

Nothing in the output indicates it. `clog()` reports a positive log-likelihood
and no warning is raised.

## Why it matters

A censored response is the ordinary encoding for below-detection-limit data,
which is common in the environmental monitoring datasets this package is used
on, and `cnorm()` on `log(y)` is the censored log-normal such data is usually
fitted with. `generate_model_set()` and `fit_model_set()` otherwise handle a
two-column censored response correctly, so a user reaches a complete, plausible
model set with no indication that its ranking is not the one the criterion
names.

Three consequences, in decreasing order of how much they change what is
reported.

- **The selected model can be wrong, not only the differences.** Selection is
  pushed toward the simpler model, because the likelihood differences are
  compressed while the `2k` penalty is not.
- **Variable importance is built from `wi.AICc`**, so it is distorted by the
  same compression. In the reproduction below a pure-noise predictor outranks a
  real one.
- **`delta.AICc <= 5` is not the threshold it appears to be.** Under `cnorm` a
  nominal 5 corresponds to roughly 45 on a correct scale, so the top-model set
  is over-permissive.

## Evidence

Simulated data with a known answer: `n = 300`, response
`1 + sin(2*pi*x) + 0.35*z` with normal errors of scale 0.5, the lowest 20 per
cent left-censored; four candidate predictors `x`, `z`, `w`, `v` of which `w`
and `v` are noise; `max.predictors = 2`, `k = 5`, eleven candidates. Measured on
[...]

PR #48, opened 2026-09-05
Rank a model set on a usable log-likelihood (#42, #44)

beckyfisher:
## What

`AICc` and `BIC` are read from `MuMIn::AICc()` and `stats::BIC()`, which both
resolve to the fitted family's own `aic` slot. Two kinds of `test.fit` break
that route, in opposite directions, and neither was reported. This branch
resolves once per model set, before any candidate is fitted, which
log-likelihood the set is ranked on.

- A **quasi-likelihood** has no log-likelihood, so the criterion is `NA` for
  every candidate. `fit_model_set()` now stops before any candidate is fitted,
  naming the family (FSSgam_package#44).
- **`mgcv`'s censored families** `cnorm()` and `clog()` return a number that is
  not built from a censored log-likelihood. The set is now ranked on a censored
  log-likelihood computed here, and a message says so (FSSgam_package#42).

`fit_model_set()`, `full_subsets_gam()` and `extract_mod_dat()` gain
`logLik.fn`, a function of one fitted model returning a single log-likelihood
value. Where supplied it replaces the log-likelihood alone: the criterion is
built at the degrees of freedom and sample size the default route uses, so every
model in the set is scored the same way.

## Why it matters

Ranking the candidate set is what this package does, and both failures reach the
whole of the output — `delta.AICc`, `delta.BIC`, both weight columns, the
top-model set and variable importance.

Neither named its cause. A quasi-likelihood set fitted successfully and
returned a table of `NA`, the only sign being four `no non-missing arguments to
min` warnings, two from `compute_model_weights()` and two from `wi()`. A censored set returned a complete, plausible ranking
that was not the one the criterion names. `quasipoisson()` and
`quasibinomial()` are ordinary choices for overdispersed counts and proportions,
and a censored response is the ordinary encoding for below-detection-limit data,
so neither is a rare case in this package's use.

## Evidence

Measured on Debian WSL2, R 4.6.1, mgcv 1.9-4, MuMIn 1.48.19, testthat 3.3.2, at
the branch head.

On the simulated set FSSgam_package#42 reports — `n = 300`, response
`1 + sin(2*pi*x) + 0.35*z`, the lowest 20 per cent left censored, four candidate
predictors of which `w` and `v` are noise, eleven candidates:

| | previous criterion | this branch |
|---|---|---|
| best model | `x` | `x+z`, the generating model |
| spread of `delta.AICc` | 35.811 | 310.773 |
| models within `delta.AICc <= 5` | 4 | 2 |
| summed importance, `z` (in the generating model) | 0.182 | 0.838 |
| summed importance, `v` (noise) | 0.253 | 0.035 |

Those are the values the issue gives for the correct criterion.

A `quasipoisson()` `test.fit` on `case_study1` now stops before fitting rather
than returning an all-`NA` criterion column, and the same set is fitted and
ranked when a log-likelihood is supplied through `logLik.fn`.

Test suite: **902 passing, 0 failing, 8 skipped blocks**, against 772 and 7 on
`dev` at `023cd1a`. `devtools::check()` in a plain clone: 0 errors, 0 warnings,
0 notes.

<details>

### Files

- **`R/criterion.R`** (new). `fit_family_name()`, `is_censored_family()`,
  `censored_loglik()`, `log1mexp()`, `criterion_from_loglik()`,
  `usable_scalar()` and `resolve_criterion()`. Nothing here is exported.
- **`R/functions_supporting.R`**. `extract_mod_dat()` gains `logLik.fn = NULL`.
  It is exported, so the argument has a default and the existing call is
  unchanged.
- **`R/fit-model-set.R`**. `fit_model_set()` gains `logLik.fn = NULL`, calls
  `resolve_criterion()` before the fitting loop, and threads the result through
  both fitting helpers to all three `extract_mod_dat()` call sites.
- **`R/full-subsets-gam.R`**. Passes it through.
- **`tests/testthat/test-criterion.R`** (new), plus fixtures in
  `helper-fixtures.R`. `test-fit_model_set_options.R`'s two
  `normalise_mod_dat_rows()` tests are rewritten against the row shape the
  unsaved loop now collects, which gained the `fit.ok` flag.
- **`DESCRIPTION`**. The `testthat` `Suggests` floor rises to 3.2.0, for
  `local_mocked_bindings()`.
- **`prompts/criterion-columns.md`** (new), per the repository's Section 7.

### Decisions

**The corrected values overwrite `AICc` and `BIC`.** The returned shape is
unchanged and every downstream name is stable, so `delta.AICc`, the weights and
variable importance are corrected with them. Adding columns would have left the
ranking the function returns wrong unless the user re-derived it.

**The censored log-likelihood is applied automatically, with one message per
model set.** A user who never reads FSSgam_package#42 gets a correct ranking;
the message makes the change in results between versions visible rather than
silent. Supplying `logLik.fn` overrides it and suppresses the message.

**The refusal is in `fit_model_set()`, not `generate_model_set()`.** It is
`logLik.fn` that decides whether a `test.fit` whose family has no
log-likelihood can be used at all, and `generate_model_set()` never sees that
argument. Refusing there would have made `logLik.fn` unreachable.

**Two guards sit either side of the fitting loop, and the two fitting paths
now agree on what a failed model is.** `save.model.fits = FALSE` defined a
failed model by an `NA` criterion, so a candidate that fitted and was given none
was reported in `failed.models` there and as a success on the other path. Both
now read it from whether `fit_mod_l()` returned a fitted object.
 `resolve_criterion()` puts
the built-in censored log-likelihood through the same check a supplied
`logLik.fn` is put through, so a censored response it cannot evaluate is
reported before anything is fitted rather than producing an `NA` criterion
column. `fit_model_set()` then stops where every candidate fitted and every one
was recorded with an `NA` criterion, which covers a `logLik.fn` that succeeds on
the `test.fit` and fails on the candidates.

**A quasi-likelihood is refused whichever route supplies it, and through
`MuMIn::uGamm()`/`mgcv::gamm()` that is a behaviour change beyond the `NA`
case**: 1.1.0 returned a complete ranked table for such a set, built from the
PQL working model. `NEWS.md` records the measured before and after.

**Two checks, because neither covers the other.** The value `MuMIn::AICc()`
returns for the `test.fit` is the general one: it does not depend on a list of
family names, so any family whose `aic` slot is undefined reaches it. The family
name is checked as well, because a quasi-likelihood fitted through
`MuMIn::uGamm()` or `mgcv::gamm()` passes the first — `AICc` for such a fit is
read from the internal `lme` fit of the PQL working model and is a number.
Measured on `case_study1`, `uGamm(..., family = quasipoisson())`: 176.03, and a
three-candidate set ranked with a weight of 1.000 on the best. A PQL
log-likelihood is a function of the working response, which itself changes with
the fixed effects, so those differences rank nothing. The name is read through
`fit_family_name()`, because `stats::family()` errors on a
`uGamm(lme4 = FALSE)` fit.

**Quasi-AIC is not offered.** `MuMIn::QAIC()`/`QAICc()` need a dispersion
estimate and a decision about which fit supplies it, which is a modelling
decision rather than a defect fix. `logLik.fn` is what a user who wants one
would build it from.

### The censored log-likelihood

`censored_loglik()` computes from the fitted mean, the scale and the response's
censoring coding, not from the family's `aic` slot, so it is independent of the
three defects FSSgam_package#42 documents and stays correct if `mgcv` repairs
them. Left, right and interval censoring, either column order, and non-unit prior
weights are all covered, and every quantity is computed in logs so that an
observation far into a tail does not underflow. The issue's sketch guarded
against right censoring and non-unit weights, and both turned out to be well
defined.

Measured on Debian WSL2, R 4.6.1, `mgcv` 1.9-4, `MuMIn` 1.48.19:

- **An extended family reports its estimated parameter as part of its name once
  fitted.** `fit$family$family` is `"cnorm(0.573)"` on the test suite's
  fixture, which is the issue's own simulated set, not `"cnorm"`, so the test is
  on the prefix. Issue FSSgam_package#42's reproduction script has the
  complementary defect: it tests `identical(family, "clog")`, which is never
  true after fitting, so it evaluated the normal density on a logistic fit and
  the `clog` figure in its table is wrong.
- **`getTheta(TRUE)` returns the scale itself, not its logarithm**, despite
  `theta` being documented as a log scale parameter. A prior weight rescales the
  family's log scale by `-log(wt)/2`, so it divides the scale by `sqrt(wt)`.
- **`attr(logLik(fit), "df")` is finite even for a quasi-likelihood fit**, whose
  `logLik()` value is `NA`: `mgcv:::logLik.gam` computes the degrees of freedom
  before it reads the `aic` slot. That is what makes `logLik.fn` a usable route
  for a quasi `test.fit`.
- The degrees of freedom convention is `sum(edf2) + scale.estimated +
  family$n.theta`, which on the suite's `cnorm` fixture is 5.996 where
  `sum(edf) + scale.estimated` is 4.919, so it is read from the fit rather than
  recomputed.

### Tests

`censored_loglik()` is checked against `censored_loglik_oracle()`, a second
computation built from `mgcv`'s own saturated log-likelihood and deviance
residuals via `deviance = 2 * (saturated - log-likelihood)`. The two share no
code: one evaluates densities and tail probabilities, the other reads the `ls`
and `dev.resids` slots, which drive REML smoothing-parameter selection and are
not among the parts FSSgam_package#42 reports as defective. They agree to 1e-8
for `cnorm` and `clog`, over left, right, interval and no censoring, and with
non-unit prior weights.

`criterion_from_loglik()` is checked to reproduce `MuMIn::AICc()` and
`stats::BIC()` exactly when handed the log-likelihood those functions use, on
Gaussian, Tweedie and negative binomial fits — the condition for the claim that
only the log-likelihood changes.

Regression tests cover the censoring codings that returned no number before:
an interval pair written in the reverse column order, which `mgcv` fits
identically; an interval far into a tail, which underflowed; and an infinite
bound written in the first column, which is refused because `mgcv` drops such a
row from the fit without a message. Two more check that the built-in is refused
up front where it cannot be evaluated, and that a candidate fitted and given no
criterion is reported the same way on both fitting paths.

No test asserts a value produced by `mgcv`'s defective `aic` slots, so a repair
upstream would not fail one; what that route produced on 2026-09-05 is recorded
in a comment beside the test instead. Three assertions on external behaviour are
deliberate and are premise checks: `MuMIn::AICc()` being `NA` for a quasi `gam`
and finite for a quasi `uGamm`, which is why two checks are needed; the
scale-recovery test, which would catch a change to `getTheta()`'s semantics; and
the `clog` coding test, which skips rather than fails if `mgcv` stops fitting
that coding.

### Two defects found while reviewing this branch, fixed here

**Summed variable importance came back `NA` for every predictor** whenever any
candidate in the table had an `NA` model weight, including predictors appearing
in no such candidate. The route to it on 1.1.0 is a candidate that failed to fit
with `save.model.fits = FALSE`, which keeps a row for it; the
`save.model.fits = TRUE` path drops the row and was unaffected, so which of the
two a user saw depended on a memory setting, and `max.models` changes that
setting without being asked. The weights are now summed with `na.rm = TRUE`, and
under the default `VI.mods` the count of models per predictor is taken over the
candidates that have a weight rather than over every row, so the two paths give
the same scores for the same model set.

**The two fitting paths disagreed on what a failed model is.**
`save.model.fits = FALSE` defined one by an `NA` criterion, so the same partial
failure returned eight successes with five `NA` rows on one path and three
successes with five `failed.models` on the other. Both now read it from whether
`fit_mod_l()` returned a fitted object, which is what the saved path has always
used.

### The parallel path

`logLik.fn` reaches a `doSNOW` worker only on the `save.model.fits = FALSE`
path, where `extract_mod_dat()` is called inside the `%dopar%`. On the saved
path the criterion is computed on the calling process after the fits return.

One restriction is documented on the argument rather than removed: under
`parallel = TRUE` with `save.model.fits = FALSE`, `foreach` replaces the
environment of a function it exports whose environment is the global one, so a
`logLik.fn` written at the top level of a script cannot see anything it refers
to and does not define. Measured in one cluster: a top-level closure reading a
top-level variable fails on the worker with "object 'shift' not found", and the
same closure built by a constructor returns the right value. This is
`beckyfisher/FSSgam#10`'s cause applied to a new argument, and it cannot be
resolved on the calling process the way that one was, because the function has
to be evaluated on the fit. The guard added here turns what would have been a
silent all-`NA` criterion column into a refusal.

Verified with a single-cluster script in a fresh process, against the installed
package (FSSgam_package#14 makes a multi-cluster script stall with high
probability, so `test-fit_model_set_parallel.R` is not the instrument): both
paths give `AICc` identical to the sequential run for all four candidates of a
`cnorm` set, with no `NA`. The saved path completed on two of three attempts,
the third stalling in the way that issue documents. The committed parallel test
is opt-in behind `skip_unless_parallel_opt_in()`, as every other one is.

### Reproduction

`ignore/censored_criterion_reprex.R` reproduces both issues end to end. It is
in the gitignored scratch folder, so it is not part of this branch.

</details>


beckyfisher:
The branch was reviewed over five rounds and changed substantially. The body
above describes the current state; this comment records what is in it that was
not in the first version, so a reader who saw that one does not have to diff.

## Two defects in the package, not in this change, fixed here

Both were found while reviewing and both are reachable on 1.1.0. Each has its
own commit, its own regression test and its own `NEWS.md` entry.

**The two fitting paths disagreed on what a failed model is.**
`save.model.fits = FALSE` identified one by an `NA` criterion and
`save.model.fits = TRUE` by the class of the fitted object, so a candidate that
fitted and was given no criterion was reported in `failed.models` by one and as
a success by the other. Measured on the eight-model `case_study1` set with five
candidates given no criterion: three successes and five `failed.models` against
eight successes and five `NA` rows. Both paths now read it from whether
`fit_mod_l()` returned a fitted object. The unsaved fitting loop carries a
`fit.ok` flag alongside each summary row for this, read and dropped before
`mod.data.out` is assembled, so the returned table is unchanged in shape.

**Summed variable importance came back `NA` for every predictor** whenever any
candidate in the table had an `NA` model weight, including predictors appearing
in no such candidate. The route on 1.1.0 is a candidate that failed to fit under
`save.model.fits = FALSE`, which keeps a row for it, where
`save.model.fits = TRUE` drops the row and was unaffected. `max.models` forces
the first setting above 200 candidates, so which of the two a user saw was not
always their choice. The weights are now summed with `na.rm = TRUE`, which
corrects `VI.mods = "all"`, and under `VI.mods = "min.n"` the count of models
per predictor is taken over the candidates that have a weight, which corrects
that branch and makes the two paths agree.

## The quasi refusal is wider than first written

Reading only the value `MuMIn::AICc()` returns misses a quasi-likelihood
supplied through `MuMIn::uGamm()` or `mgcv::gamm()`: for such a fit that value
is read from the internal `lme` fit of the PQL working model and is a number.
Measured on `case_study1`, `uGamm(..., family = quasipoisson())` with three
candidates, 1.1.0 returned `AICc` 170.261, 176.030 and 138.582 and a weight of
1.000 on the best. A PQL log-likelihood is a function of the working response,
which itself changes with the fixed effects, so those differences rank nothing.
The fitted family's name is now checked beside the value, and such a set stops
where it previously returned a complete table. That is a behaviour change beyond
the `NA` case and is recorded in `NEWS.md`.

## The censored log-likelihood covers more codings than the issue's sketch

`mgcv` orders a finite interval's two columns with `pmin`/`pmax`, so a pair
written in either order fits identically and is now read the same way. An
infinite bound written in the first column is refused instead: `mgcv` reads a
censored case from the second column alone, so under `clog()` such a row is
dropped from the fit without a message — measured, 60 of 200 rows giving −204.99
scored against −88.42 from the fit's own deviance — and `cnorm()` does not fit
that coding at all. Every quantity is computed in logs, so an observation far
into a tail no longer underflows to `-Inf` and take the criterion with it. Right
censoring and non-unit prior weights are handled rather than guarded against, as
the issue's sketch proposed.

## Two guards, either side of the fitting loop

`resolve_criterion()` puts the built-in censored log-likelihood through the same
check a supplied `logLik.fn` is put through, so a censored response it cannot
evaluate is reported before anything is fitted. `fit_model_set()` then stops
where every candidate that fitted was given no criterion, and warns naming them
where only some were. Without the first, a censored coding the built-in could
not evaluate produced an `NA` criterion for every candidate under a message
saying the ranking had been corrected — `FSSgam_package#44`'s output reached by
another route.

## One restriction is documented rather than removed

Under `parallel = TRUE` with `save.model.fits = FALSE`, the only combination
that evaluates `logLik.fn` on a worker, `foreach` replaces the environment of an
exported function whose environment is the global one, and `.GlobalEnv` is not
serialised by value in any case. A `logLik.fn` written at the top level of a
script therefore cannot see anything it refers to and does not define. Measured
in one cluster: a top-level closure reading a top-level variable fails with
`object 'shift' not found`, and the same closure built by a constructor returns
the right value. This is `beckyfisher/FSSgam#10`'s cause reaching a new
argument, and unlike that one it cannot be resolved on the calling process,
because the function has to be evaluated on each fit. It is documented on the
argument, and the guard above turns what would have been a silent all-`NA`
criterion column into a refusal.

## Also worth knowing

`FSSgam_package#42`'s own reproduction script tests
`identical(fit$family$family, "clog")`, which is never true for a fitted
extended family — the name includes the estimate, `clog(0.331)` — so it
evaluated the normal density on a logistic fit and the `clog` row of that
issue's table is wrong. The correct value on that fit is −250.82, which this
branch returns and which agrees with `mgcv`'s own `ls`/`dev.resids` to 6e-14.
The branch matches on the prefix for this reason.

`CLAUDE.md` gains a Phase 15 section holding what a later session would
otherwise re-derive: the three `mgcv` properties the implementation depends on,
the prefix-match rule, the censored column conventions, the worker restriction,
and why the test oracle must stay independent of the implementation.

## Verification at the branch head

Debian WSL2, R 4.6.1, `mgcv` 1.9-4, `MuMIn` 1.48.19, `testthat` 3.3.2.
`testthat::test_local()`: 902 passing, 0 failing, 8 skipped blocks, against 772
and 7 on `dev` at `023cd1a`. `devtools::check()` in a plain clone: 0 errors, 0
warnings, 0 notes. All nine CI checks pass, `parallel-tests` included. Both
fitting paths were verified under `parallel = TRUE` one cluster per fresh
process against an installed copy, reproducing the sequential `AICc` exactly;
the saved path completed on two of three attempts, the third stalling in the way
`FSSgam_package#14` documents.

